PR #5, opened 2021-03-04
Barebones GitHub Actions CI pipeline

ali-ramadhan:


tomchor (approved):

ali-ramadhan:
Looks like tests pass! Although I guess it only tests whether `using Oceanostics` work haha.

I added some `[compat]` entires as a start so we should be able to register as a package after merging.

I also added CompatHelper which will open PRs updating `[compat]` as other packages tag new versions. And TagBot which will automate the tagging process and add useful release notes to each tag.

tomchor (commented):
I wonder if it's a bad idea to change the Oceananigans compat to "0" (if that even works).

The rationale being that Oceananigans bumps versions so fast (and most of those changes an unimportant to this package per se) that it'll simply prevent us from needing to update `compat` on a weekly basis.

ali-ramadhan:
Ah true, that sounds like a good idea.

You might have to use `"^0"` or `"~0"` (https://julialang.github.io/Pkg.jl/dev/compatibility/#Caret-specifiers).

I think the registrator failed on me at some point because of the "0" compat but worth trying, I might be wrong!

tomchor:
You're probably right, we need to have an upper bound. We can merge this for now and later on I'll try to use "^0".
Feel free to merge and I'll try to register the package.

PR #9, opened 2021-04-09
Changed order of operations in kernel for TKE dissipation rate

tomchor:
@glwagner I think we discussed this at some point. Interpolating a quantity and then squaring it is different from squaring it and then interpolating (and I think the latter is preferred). CC'ing you because I think you might be using a kernel for ε that [interpolates and then squares](https://github.com/CliMA/LESbrary.jl/blob/cf31b0ec20219d5ad698af334811d448c27213b0/src/TurbulenceStatistics/viscous_dissipation.jl#L52-L64).

Just as a reference I'm plotting the different between both for a simulation of mine. I don't think the difference is negligible:

Screenshot from 2021-04-09 09-54-42

(I'd expect the square->interpolate values to be generally smaller than the interpolate->square values, but I'm not 100% sure that should be the case)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceanostics"
 uuid = "d0ccf422-c8fb-49b5-a76d-74acdde946ac"
 authors = ["tomchor <tomaschor@gmail.com>"]
-version = "0.2.1"
+version = "0.3.0"
 
 [deps]
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
--- a/src/TurbulentKineticEnergyTerms.jl
+++ b/src/TurbulentKineticEnergyTerms.jl
@@ -58,9 +58,27 @@ end
 
 
 
-
 @inline fψ²(i, j, k, grid, f, ψ) = @inbounds f(i, j, k, grid, ψ)^2
-@kernel function anisotropic_viscous_dissipation_rate_ccc!(ϵ, grid, u, v, w, params)
+@kernel function isotropic_pseudo_viscous_dissipation_rate_ccc!(ϵ, grid, u, v, w, ν)
+    i, j, k = @index(Global, NTuple)
+
+    ddx² = ∂xᶜᵃᵃ(i, j, k, grid, ψ², u) + ℑxyᶜᶜᵃ(i, j, k, grid, fψ², ∂xᶠᵃᵃ, v) + ℑxzᶜᵃᶜ(i, j, k, grid, fψ², ∂xᶠᵃᵃ, w)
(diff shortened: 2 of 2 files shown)


glwagner:
Huh, that's interesting. I haven't worked with the dissipation function myself but @qingli411 might be interested or having something to add.

If we want to isolate the squared quantities and compute them prior to interpolating, shouldn't we end up with cross terms? For example,

```
Sigma_xy^2 = 1/4 * (u_y + v_z)^2 = 1/4 * u_y^2 + 1/2 * u_y * v_z + 1/4 * v_z^2
```

Or have the cross terms been eliminated somehow?

glwagner:
@jm-c has mentioned some rules of thumb for computations (but mostly regarding prognostic calculations, not diagnostic).

I suppose to figure out the "true" dissipation, we need to do a discrete calculation based on a derivation of discrete TKE from the discrete momentum equations? In deriving a "discrete TKE equation" we have to perform interpolation; but I suppose the order in which we perform interpolation might affect how we are able to combine / cancel out terms to obtain the TKE equation we expect to see. I'm not sure if anyone has done this.

tomchor:
>  Sigma_xy^2 = 1/4 * (u_y + v_z)^2 = 1/4 * u_y^2 + 1/2 * u_y * v_z + 1/4 * v_z^2

Yeah, in the case of your Σ formulation I think the answer is less clear. I just thought I should let you know. The formulation I'm using is a more traditional implementation of `(du_i / dx_j)^2` where I just sum up all the tensor coordinates one by one. Any particular reason you guys chose the Σ form?

glwagner:
That is the form the dissipation rate takes when using an LES closure formulated in terms of the strain rate.

glwagner:
Or perhaps more precisely we find that in the continuous equations the dissipation rate density (the volume integral of which becomes the total dissipation rate) is

```
ep = - 2 * nu_e * Sigma_ij^2
```

One reason we choose this form is because it shows that the dissipation rate is positive-definite.

I tried to puzzle out a way to write it solely in terms of `d_i u_j` but I don't quite see it. I do think you might get this form when the viscous operator is written a bit differently, eg as `d_j nu * u_{i, j}`. 

This raises a separate question on whether we should really implement `IsotropicDiffusivity` in Oceananigans in terms of the rate-of-strain. I suspect we should... but perhaps you have an opinion about this @tomchor (and we can discuss this over at Oceananigans.jl if need be).

tomchor:
@glwagner So, just to complement our discussion yesterday. The viscous dissipation rate is always

`ε = 2 ν S_ij S_ij`,
however in the case of a homogeneous flow, this can be equivalent to 

`ε* = ν (∂_j u_i)^2`
which is called pseudo-dissipation (see Pope's book). Under normal circumstances they differ by about 2% only. In any case, I corrected the nomenclature in Oceanostics to include the word "pseudo" in order to be more accurate. 

Although I haven't gone through the discretized equations, I'm following the the of thumb and squaring quantities first and interpolating only in the last step. This is easy to do with the pseudo-dissipation, but not trivial for the actual dissipation though...

tomchor:
@JuliaRegistrator register

tomchor (commented):
@JuliaRegistrator register

PR #20, opened 2021-05-28
Renamed TurbulentKineticEnergyTerms to TKEBudgetTerms

tomchor:
Renamed this module because to be more accurate. TKE (`(u^2+v^2+w^2)/2`) only has three terms. These are the terms for the TKE equation.

I just wanted to some feedback to see if this is better for other people.

CC @glwagner @ali-ramadhan


glwagner:
Could also use "Budget" rather than "Equation" (same thing, but possibly slightly clearer)

glwagner (approved):
Wholeheartedly approve.

The name of the PR is outdated; might be nice to update before merging so its correct in history.

PR #22, opened 2021-05-28
Modified terms from ShearProd_x to XShearProd to be more consistent with Oceananigans

### LINKED ISSUE #19: TitleCase names, plus English language conventions for written variable names?
There are some quantities like `ShearProduction_x` that blend `TitleCase` and `snake_case` styles:

https://github.com/tomchor/Oceanostics.jl/blob/c7c6a8bea46d127f243e7c8f610edb7ce03ffcec/src/TurbulentKineticEnergyTerms.jl#L226

It's not a big deal, but I think it's best to "practice what we preach", if we are going to ask future developers to conform to style guidelines.

The other little inconsistency is that this name "blends" English-language conventions (where descriptors come first) with a "subscripting" convention (eg `_x` refers to the x-component). For example, in Oceananigans we use `ZDirection` rather than `Direction_z`.

We could call this instead `XShearProduction`; also `YShearProduction` and `ZShearProduction`.

This is also more consistent with some `Oceananigans` variables like `XFaceField` or `ZDirection`.

tomchor:
Closes #19 
cc @glwagner

### DIFF
--- a/src/TurbulentKineticEnergyTerms.jl
+++ b/src/TurbulentKineticEnergyTerms.jl
@@ -3,8 +3,8 @@ module TurbulentKineticEnergyTerms
 export TurbulentKineticEnergy, KineticEnergy
 export IsotropicViscousDissipationRate, IsotropicPseudoViscousDissipationRate
 export AnisotropicViscousDissipationRate, AnisotropicPseudoViscousDissipationRate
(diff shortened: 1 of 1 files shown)


glwagner (approved):
🎉

Issue #21, opened 2021-05-28
Change every `KernelComputedField` to `KernelFunctionOperation`


tomchor:
@glwagner do you see any problems downsides for this? If I understand correctly the migration will only bring upsides:

- Simpler (no need for `@kernel` and `@index`)
- KernelAbstrations won't be a dependence anymore
- Resulting quantities are abstract operations, which are more flexible

glwagner:
I don't see any downsides. You'll only need `KernelComputedField` to do fairly non-trivial things in a kernel like unroll loops, etc.

tomchor:
Good. Then I'll probably start some migration soon.

I haven't used `KernelFunctionOperation` so I'm not familiar, but I guess if we use user-facing constructors they would return the operation itself, right? (They wouldn't return, for example, a `ComputedField` based on the operation, right?)

glwagner:
Hrmm yeah I think the minimal / easy to maintain API returns the operations and let users build `ComputedField` or `AveragedField` according to their needs.

PR #31, opened 2021-08-18
Migrate quantities from KernelComputedFields to AbstractOperations

tomchor:
The idea is that 

- Quantities that can be represented with `AbstractOperations` that can currently be compiled on GPUs should become simple `AbstractOperations`
- Quantities that are too complex to be represented with `AbstractOperations`  on GPUs (like TKE dissipation rate) become `KernelFunctionOperations`

Started this with `FlowDiagnostics`.

Closes https://github.com/tomchor/Oceanostics.jl/issues/21

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -26,47 +26,50 @@ function RichardsonNumber(model; N²_bg=0, dUdz_bg=0, dVdz_bg=0)
 end
 
 
-function RossbyNumber(model; dUdy_bg=0, dVdx_bg=0)
+function RossbyNumber(model; dUdy_bg=0, dVdx_bg=0, f=nothing)
     u, v, w = model.velocities
-    f₀ = model.coriolis.f
+    if f==nothing
+        f = model.coriolis.f
+    end
 
     dUdy_tot = ∂y(u) + dUdy_bg
     dVdx_tot = ∂x(v) + dVdx_bg
 
-    return (dVdx_tot - dUdy_tot) / f₀
+    return (dVdx_tot - dUdy_tot) / f
 end
 
 
 
 #++++ Potential vorticity
-function potential_vorticity_in_thermal_wind_fff(i, j, k, grid, u, v, b, f₀)
-    i, j, k = @index(Global, NTuple)
+function potential_vorticity_in_thermal_wind_fff(i, j, k, grid, u, v, b, f)
 
     dVdx =  ℑzᵃᵃᶠ(i, j, k, grid, ∂xᶠᵃᵃ, v) # F, F, C → F, F, F
     dUdy =  ℑzᵃᵃᶠ(i, j, k, grid, ∂yᵃᶠᵃ, u) # F, F, C → F, F, F
     dbdz = ℑxyᶠᶠᵃ(i, j, k, grid, ∂zᵃᵃᶠ, b) # C, C, F → F, F, F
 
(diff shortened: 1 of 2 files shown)


src/FlowDiagnostics.jl
```diff
@@ -1,35 +1,45 @@
...
-    @inbounds Ri[i, j, k] = dBdz_tot / (dUdz_tot^2 + dVdz_tot^2)
+    return dBdz_tot / (dUdz_tot^2 + dVdz_tot^2)
 end
 
 
-@kernel function rossby_number_ffc!(Ro, grid, u, v, params)
-    i, j, k = @index(Global, NTuple)
+function RossbyNumber(model; dUdy_bg=0, dVdx_bg=0)
+    u, v, w = model.velocities
+    f₀ = model.coriolis.f
```
glwagner, line 31: This only works with `FPlane`. Is that intended?

Alternatively this object could take `u, v, f` as arguments rather than `model`; then users can provide correct values as needed, and because we're using abstract operations it might even work with something like

```
f(x, y, z) = f0 + beta * y
```

since we can use functions in AbstractOperations.

src/FlowDiagnostics.jl
```diff
@@ -43,12 +53,20 @@ end
...
-    @inbounds PV[i, j, k] = pv_barot[i, j, k] + pv_baroc[i, j, k]
+    return pv_barot + pv_baroc
+end
+
+function ThermalWindPotentialVorticityᶠᶠᶠ(model)
+    u, v, w = model.velocities
+    b = model.tracers.b
+    f₀ = model.coriolis.f
+    return KernelFunctionOperation{Face, Face, Face}(potential_vorticity_in_thermal_wind_fff, model.grid;
+                         computed_dependencies=(u, v, w, b), parameters=f₀)
```
glwagner, line 64: indenting is off here

src/FlowDiagnostics.jl
```diff
@@ -43,12 +53,20 @@ end
...
     pv_baroc = -f₀*(dUdz^2 + dVdz^2)
 
-    @inbounds PV[i, j, k] = pv_barot[i, j, k] + pv_baroc[i, j, k]
+    return pv_barot + pv_baroc
+end
+
+function ThermalWindPotentialVorticityᶠᶠᶠ(model)
+    u, v, w = model.velocities
+    b = model.tracers.b
+    f₀ = model.coriolis.f
```
glwagner, line 62: same comment as above; this only works with `model.coriolis isa FPlane`

glwagner (approved):
Seems like a good idea. Left a few comments that could be addressed or saved for later.

src/FlowDiagnostics.jl
tomchor: Yeah it only works with f-plane. This was only because I didn't think to pass functions to `AbstractOperations`!

But I think your idea is good and it appears to work (I just tested it). The caveat is that this makes it more likely that this won't compile in GPUs depending on the function

src/FlowDiagnostics.jl
tomchor: DOne! Thanks

src/FlowDiagnostics.jl
tomchor: GPU seems to be able to compile at least reasonable functions, so  adopted your suggestion

tomchor:
Addressed the comments. If you're happy with it and the tests pass I'll merge and to the other migrations to KernelFunctionOp in another PR

Issue #44, opened 2021-09-06
Making operators more general


whitleyv:
I have noticed that some of the derivative operators only work for the `Rectilinear Grid` in Oceananigans. Such as this one in the PV diagnostic:
https://github.com/tomchor/Oceanostics.jl/blob/4daaeb67e3c017560af8dd1d13aa902c28ca3251/src/FlowDiagnostics.jl#L74

I believe if you use  `∂yᶜᶠᵃ(i, j, k, grid, u, args...)` operator here, it would be more general on the grid requirement, but would still allow for the location you want. Probably not a pressing problem, as not many people would be using something other than rectilinear. I'm trying to use these with an ImmersedGrid, which, in this aspect, should work the same as the rectilinear, but it has a different name.

tomchor:
You're right! I think no one has used Oceanostics with a grid that isn't a `RegularRectilinearGrid`, so I haven't thought much about that. I've never used operators such as `∂yᶜᶠᵃ`. Do they dispatch the same as `∂yᵃᶠᵃ` if the grid is `RegularRectilinear`?

Also, sorry for the naive question but what's different between an `ImmersedGrid` and a `RegularRectilinearGrid`? I haven't been exposed to it.

If you're using these and need them to be more general, feel free to open a PR to modify just the diagnostics you want. We don't need to migrate them all at once I think. I've been making changes little by little as time permits :)

glwagner:
> You're right! I think no one has used Oceanostics with a grid that isn't a `RegularRectilinearGrid`, so I haven't thought much about that. I've never used operators such as `∂yᶜᶠᵃ`. Do they dispatch the same as `∂yᵃᶠᵃ` if the grid is `RegularRectilinear`?
> 
> Also, sorry for the naive question but what's different between an `ImmersedGrid` and a `RegularRectilinearGrid`? I haven't been exposed to it.
> 
> If you're using these and need them to be more general, feel free to open a PR to modify just the diagnostics you want. We don't need to migrate them all at once I think. I've been making changes little by little as time permits :)

`ImmersedBoundaryGrid` is a wrapper that modifies an "underlying grid" to account for the presence of a solid boundary immersed within the grid.

It's probably best for future-proofing to always use the "two horizontal location" operators (such as `∂yᶜᶠᵃ` where applicable. This way code is valid on curvilinear grids in addition to rectilinear grids.

[...]

PR #45, opened 2021-09-08
Generalize derivatives in PV diagnostic

whitleyv:
As per issue  #44, this PR generalizes the Ertel PV and Thermal Wind PV derivative inputs to allow for any grid types by replacing the x and y derivatives that only dispatch on `AbstractRectilinearGrid` in Oceananigans. Instead of specifying one location, ∂yᵃᶠᵃ, both horizontal locators are used, such as in ∂yᶜᶠᵃ. This does not impact the vertical derivatives.


tomchor (approved):

tomchor:
Thanks @whitleyv!

PR #50, opened 2021-11-02
Fixes TimedProgressMessenger's CFL display

tomchor:
For some reason output was always 0 before this change

### DIFF
--- a/src/progress_messengers.jl
+++ b/src/progress_messengers.jl
@@ -110,11 +110,9 @@ end
 
 
 
-mutable struct TimedProgressMessenger{T, A, D, L, Δ} <: Function
+mutable struct TimedProgressMessenger{T, L, Δ} <: Function
     wall_time₀ :: T  # Wall time at simulation start
     wall_time⁻ :: T  # Wall time at previous calback
-       adv_cfl :: A
-       dif_cfl :: D
            LES :: L
             Δt :: Δ
 end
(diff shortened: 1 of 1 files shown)


glwagner (approved):
It looks like maybe the CFL objects were built with `dt=0`, and then `dt` was not updated? This could warrant an improvement in the UI for the CFL objects. But it's not urgent. Probably what would really be useful for users is a function like

```julia
cfl(simulation) = # calculate cfl
```

and maybe the more specific versions

```julia
advective_cfl(simulation) = ...
diffusive_cfl(simulation) = ...
```

Then users don't have to create the objects `AdvectiveCFL`, etc.

Another comment which is unrelated to this PR is that I think "LES` really means "variable diffusivity". There are models that most people might not call "LES" which still have a variable diffusivity.

tomchor:
> It looks like maybe the CFL objects were built with `dt=0`, and then `dt` was not updated? This could warrant an improvement in the UI for the CFL objects. But it's not urgent. Probably what would really be useful for users is a function like
> 
> ```julia
> cfl(simulation) = # calculate cfl
> ```
> 
> and maybe the more specific versions
> 
> ```julia
> advective_cfl(simulation) = ...
> diffusive_cfl(simulation) = ...
> ```
> 
> Then users don't have to create the objects `AdvectiveCFL`, etc.
> 
> Another comment which is unrelated to this PR is that I think "LES` really means "variable diffusivity". There are models that most people might not call "LES" which still have a variable diffusivity.

I agree on all fronts.

About the LES nomenclature: I couldn't find a better name that was more accurate. Naming the option `variable_diffusivity`, or `eddy_diffusivity` didn't seem like it would be user-friendly to me. Do you have suggestions?

I get we could also inspect `typeof(model.closure)` and choose appropriately inside the progress messenger function, scrapping the `LES` option altogether...

glwagner:
Maybe a kwarg that allows users to pass fields for which they'd like to see statistics? eg 

```julia
messenger = TimedProgressMessenger(summarize=model.diffusivity_fields)
```

(Or any other named tuple). Then users can pass diffusivity fields (and an example in the docs can illustrate this) or named tuples of fields. We could start by just printing the maximum value of the fields as a "summary", but could expand d in the future to min, max, mean, or whatever. That'd be easy to support, and is flexible.

PR #52, opened 2021-11-04
Fixes behavior related to Callbacks

tomchor:


tomchor:
First commit fixes the construction of the simulation in the tests. Now there's no need to figure out how to compute the time per time step:

https://github.com/tomchor/Oceanostics.jl/blob/99c34e84432859647ab290449bc085d38f2c3fa2/src/progress_messengers.jl#L149

src/progress_messengers.jl
```diff
@@ -5,16 +5,13 @@ using Printf
...
-get_Δt(wizard::TimeStepWizard) = wizard.Δt
-
 
 function SimpleProgressMessenger_function(simulation; LES=false, SI_units=true,
                                            initial_wall_time_seconds=1e-9*time_ns())
     model = simulation.model
     Δt = simulation.Δt
 
-    i, t = model.clock.iteration, model.clock.time
+    iter, t = model.clock.iteration, model.clock.time
```
glwagner, line 14:
```suggestion
    iter, t = iteration(simulation), time(simulation)
```

glwagner: and more below.

glwagner (approved):

Issue #37, opened 2021-08-20
Making buoyancy calculations more general


tomchor:
At the moment all the diagnostics that include buoyancy (like PV and Ri) only work if `buoyancy_model=BuoyancyTracer()`, but at some point it would be good to make the code more general so as to include `SeawaterBuoyancy()` as well.

PR #54, opened 2021-11-12
Allow user to pass a buoyancy field to `RichardsonNumber`

ali-ramadhan:
Right now `RichardsonNumber` cannot be used for models with `SeawaterBuoyancy` as it assumes `model.tracers.b` exists.

This PR allows users to pass in a `b` so that `BuoyancyField(model)` can be passed in. In particular I'm trying to revive https://github.com/CliMA/LESbrary.jl/blob/master/examples/three_layer_constant_fluxes.jl

Hoping to tag v0.6.2 once this is merged if that's okay.

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -15,9 +15,8 @@ using Oceananigans.Grids: Center, Face
 @inline fψ²(i, j, k, grid, f, ψ) = @inbounds f(i, j, k, grid, ψ)^2
 
 
-function RichardsonNumber(model; N²_bg=0, dUdz_bg=0, dVdz_bg=0)
+function RichardsonNumber(model; b=model.tracers.b, N²_bg=0, dUdz_bg=0, dVdz_bg=0)
     u, v, w = model.velocities
-    b = model.tracers.b
 
     dBdz_tot = ∂z(b) + N²_bg
     dUdz_tot = ∂z(u) + dUdz_bg


src/FlowDiagnostics.jl
```diff
@@ -15,9 +15,8 @@ using Oceananigans.Grids: Center, Face
 @inline fψ²(i, j, k, grid, f, ψ) = @inbounds f(i, j, k, grid, ψ)^2
 
 
-function RichardsonNumber(model; N²_bg=0, dUdz_bg=0, dVdz_bg=0)
+function RichardsonNumber(model; b=model.tracers.b, N²_bg=0, dUdz_bg=0, dVdz_bg=0)
```
tomchor, line 18: Oh, wow, I'm surprised this works. This sort of statement doesn't work in python :+1:

tomchor (approved):

tomchor:
> Hoping to tag v0.6.2 once this is merged if that's okay.

Tag away!

src/FlowDiagnostics.jl
ali-ramadhan: Ah haha yeah I think it only works for kwargs but still pretty nice.

tomchor:
One comment is that I think in the future we want to make the calculation of N2 more general, as per https://github.com/tomchor/Oceanostics.jl/issues/37

Then we could use Oceananigans buoyancy function to get N2 from both `BuoyancyTracer` and `SeawaterBuoyancy` models directly, with no need to pass a buoyancy field specificaly.

ali-ramadhan:
Ah true, forgot to check issues to still if this was one!

tomchor:
> Ah true, forgot to check issues to still if this was one!

All good. That can be done in a later PR. This is already an improvement, so I'm def okay with merging this and deealing with the rest later :)

PR #60, opened 2022-01-19
Updates Oceanostics to Oceananigans 0.68

glwagner:
This PR updates Oceanostics tests to Oceananigans 0.68, which has new syntax for computed and averaged fields.


tomchor:
Thanks, @glwagner, but I think you missed that I had already done this work and requested your review here: https://github.com/tomchor/Oceanostics.jl/pull/59

I prefer merging #59 since I changed some things that this PR doesn't (the README for example), but we can discuss. The main difference in both is the compat entry and the tagged version.

PR #61, opened 2022-01-19
Use `b=BuoyancyField(model)` instead of `model.tracers.b`

### LINKED ISSUE #37: Making buoyancy calculations more general

glwagner:
This generalizes to `SeawaterBuoyancy`.

Closes #37

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -15,7 +15,7 @@ using Oceananigans.Grids: Center, Face
 @inline fψ²(i, j, k, grid, f, ψ) = @inbounds f(i, j, k, grid, ψ)^2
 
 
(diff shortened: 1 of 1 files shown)


tomchor:
Nice. Is this enough already to clsoe https://github.com/tomchor/Oceanostics.jl/issues/37 ?

glwagner:
Yes! Only caveat is that it won't work with `buoyancy=nothing` until https://github.com/CliMA/Oceananigans.jl/pull/2155.

Also if we make some changes suggested in https://github.com/CliMA/Oceananigans.jl/issues/2022 then we probably want to use `b = buoyancy(model)` (which will avoid extra calculations when possible).

tomchor (approved):
LGTM

Issue #64, opened 2022-01-31
Should we change the way compat with Ocenanigans is dealt with?


tomchor:
At the moment the compat entry for Oceananigans is done like this: https://github.com/tomchor/Oceanostics.jl/blob/5c4288cf84f49130fbbf1c409e24ce1082a18d37/Project.toml#L10-L11

SInce Oceananigans releases minor versions so fast though (much more often than Oceanostics) generally the `compat` entry has to be updated fairly often, but it'll take a while for that `compat` update to be included in the tagged release. This means that if a user has the most up-to-date Oceananigans and tries to install Oceanostics with `] add Oceanostics`, often Oceananigans gets downgraded. For example in the example below, since the tagged version only works with Ocenanigans 0.68:

```julia
(AGUOSM22-LES-tutorial) pkg> add Oceanostics
    Updating registry at `~/.julia/registries/General`
   Resolving package versions...
   Installed Oceanostics ── v0.6.3
   Installed Plots ──────── v1.25.7
   Installed Oceananigans ─ v0.68.7
    Updating `~/repos/AGUOSM22-LES-tutorial/Project.toml`
  [033835bb] ↑ JLD2 v0.4.15 ⇒ v0.4.19
  [9e8cae18] ↓ Oceananigans v0.69.2 ⇒ v0.68.7
  [d0ccf422] + Oceanostics v0.6.3
  [91a5bcdd] ↑ Plots v1.25.0 ⇒ v1.25.7
    Updating `~/repos/AGUOSM22-LES-tutorial/Manifest.toml`
  [28b8d3ca] ↑ GR v0.62.1 ⇒ v0.63.1
  [033835bb] ↑ JLD2 v0.4.15 ⇒ v0.4.19
  [9e8cae18] ↓ Oceananigans v0.69.2 ⇒ v0.68.7
  [d0ccf422] + Oceanostics v0.6.3
  [91a5bcdd] ↑ Plots v1.25.0 ⇒ v1.25.7
  [01d81517] ↑ RecipesPipeline v0.4.1 ⇒ v0.5.0
  [05181044] + RelocatableFolders v0.1.3
  [41fe7b60] + Unzip v0.1.2
Precompiling project...
  ✓ RecipesPipeline
  ✓ Oceananigans
  ✓ Plots
  4 dependencies successfully precompiled in 38 seconds (182 already precompiled)
  3 dependencies precompiled but different versions are currently loaded. Restart julia to access the new versions
```

[...]

PR #65, opened 2022-01-31
Updates CompatHelper to automatically tag patch versions

tomchor:
Closes https://github.com/tomchor/Oceanostics.jl/issues/64


glwagner (approved):
Let's try it out.

PR #74, opened 2022-03-23
Fixes bug in `isotropic_dissipation_rate_ccc`

glwagner:
I think we may not have a test to catch this?


glwagner:
Except there is

https://github.com/tomchor/Oceanostics.jl/blob/5dde8235478fb072ea607f905fd17f5fc2e0cddd/test/runtests.jl#L146-L147

So I don't know how this slipped through.

PR #76, opened 2022-04-04
Implements test for `make_message`

glwagner:
This PR splits message "making" and message printing into two functions so we can test that the `single_line=true` works.


tomchor (approved):
Great solution!

PR #88, opened 2022-09-22
Fixes bug in IsotropicTracerVarianceDissipationRate

tomchor:


tomchor:
The alterations done so far do fix a bug, however, this routine still isn't working because it turns out that `calc_κᶜᶜᶜ` isn't working at the moment.

I'll open an issue (and possibly a bug fix) upstream. Do we want/need to rely on that subroutine?

PR #96, opened 2022-11-30
Adds DirectionalErtelPotentialVorticity

tomchor:
This PR adds `DirectionalErtelPotentialVorticity`, which computes the contribution to the Ertel PV from a specific direction. This is useful for rotated domains, especially when studying symmetric instabilities, and it's not trivial to implement since it needs several sines and cosines that don't compile on a GPU using abstract operations.

I don't love the name `DirectionalErtelPotentialVorticity` but I couldn't find a better one. So I'm open to name suggestions here.

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -215,6 +215,10 @@ end
 function DirectionalErtelPotentialVorticity(model, direction; location = (Face, Face, Face))
     validate_location(location, "DirectionalErtelPotentialVorticity", (Face, Face, Face))
 
+    if model.buoyancy == nothing || !(model.buoyancy.model isa Oceananigans.BuoyancyTracer)
+        throw(ArgumentError("`DirectionalErtelPotentialVorticity` is only implemented for `BuoyancyTracer`"))
+    end
+
     if model isa NonhydrostaticModel
         full_fields = add_background_fields(model)
         u, v, w, b = full_fields.u, full_fields.v, full_fields.w, full_fields.b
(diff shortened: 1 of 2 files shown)


src/FlowDiagnostics.jl
```diff
@@ -187,6 +187,71 @@ function ErtelPotentialVorticity(model; location = (Face, Face, Face))
...
+    dVdx =  ℑzᵃᵃᶠ(i, j, k, grid, ∂xᶠᶠᶜ, v) # C, F, C  → F, F, C → F, F, F
+    dUdy =  ℑzᵃᵃᶠ(i, j, k, grid, ∂yᶠᶠᶜ, u) # F, C, C  → F, F, C → F, F, F
+    ω̂_z = dVdx - dUdy # F, F, F
+
+    dbdx̂ = ℑyzᵃᶠᶠ(i, j, k, grid, ∂xᶠᶜᶜ, b) # C, C, C  → F, C, C → F, F, F
+    dbdŷ = ℑxzᶠᵃᶠ(i, j, k, grid, ∂yᶜᶠᶜ, b) # C, C, C  → C, F, C → F, F, F
+    dbdẑ = ℑxyᶠᶠᵃ(i, j, k, grid, ∂zᶜᶜᶠ, b) # C, C, C  → C, C, F → F, F, F
+
+    ω_dir = ω̂_x * params.dir_x + ω̂_y * params.dir_y + ω̂_z * params.dir_z
+    dbddir = dbdx̂ * params.dir_x + dbdŷ * params.dir_y + dbdẑ * params.dir_z
```
glwagner, line 210: Are we sure this the correct discrete form? If so we should write up some docs on how we deduced it.

PR #97, opened 2022-12-01
Expands `TracerVarianceDissipationRate` to work with both AMD and Smagorinsky closure

tomchor:
At the moment `TracerVarianceDissipationRate` only works with AMD since the function `_calc_κccc` isn't defined for any other closure in the Oceananigans side (see https://github.com/CliMA/Oceananigans.jl/issues/2751).


tomchor:
@glwagner if you have an easier way to support closures other than AMD please feel free to chime in. I couldn't come up with a better way than this one, although I'm sure it exists.

glwagner:
I suggest using the `diffusivity` interface to extract the diffusion coefficient.

src/FlowDiagnostics.jl
```diff
@@ -255,12 +255,22 @@ end
...
+    return _calc_κᶜᶜᶜ(i, j, k, grid, closure, tracers, id, velocities)
+end
+
 @inline function isotropic_tracer_variance_dissipation_rate_ccc(i, j, k, grid, c, velocities, p)
     dcdx² = ℑxᶜᵃᵃ(i, j, k, grid, fψ², ∂xᶠᶜᶜ, c) # C, C, C  → F, C, C  → C, C, C
     dcdy² = ℑyᵃᶜᵃ(i, j, k, grid, fψ², ∂yᶜᶠᶜ, c) # C, C, C  → C, F, C  → C, C, C
     dcdz² = ℑzᵃᵃᶜ(i, j, k, grid, fψ², ∂zᶜᶜᶠ, c) # C, C, C  → C, C, F  → C, C, C
 
-    κ = _calc_κᶜᶜᶜ(i, j, k, grid, p.closure, c, p.id, velocities)
+    κ = _calc_κᶜᶜᶜ(i, j, k, grid, p.closure, p.buoyancy, velocities, tracers, p.id)
```
glwagner, line 273: I'm not sure this formula is the correct discrete dissipation rate formula, because `κᶜᶜᶜ` is never invoked by the solver.

glwagner:
Ok, here's one way. It should be tested, but will work (in principle) with any turbulence closure. It may only return expected results for second-order operators though.

```julia
# χ is variance dissipation
#
# There are three discrete contributions to χ, each with a different location:
#
# χ_x at fcc
# χ_y at cfc
# χ_z at ccf

@inline χxᶠᶜᶜ(i, j, k, grid, c, args...) = δxᶠᵃᵃ(i, j, k, grid, c) * diffusive_flux_x(i, j, k, grid, args...) / Vᶜᶜᶜ(i, j, k, grid)
@inline χyᶜᶠᶜ(i, j, k, grid, c, args...) = δyᵃᶠᵃ(i, j, k, grid, c) * diffusive_flux_y(i, j, k, grid, args...) / Vᶜᶜᶜ(i, j, k, grid)
@inline χzᶜᶜᶠ(i, j, k, grid, c, args...) = δzᵃᵃᶠ(i, j, k, grid, c) * diffusive_flux_z(i, j, k, grid, args...) / Vᶜᶜᶜ(i, j, k, grid)

# We can reconstruct this at cell centers:

@inline χᶜᶜᶜ(i, j, k, grid, c, args...) = ℑxᶜᵃᵃ(i, j, k, grid, χxᶠᶜᶜ, c, args...) +
                                          ℑyᵃᶜᵃ(i, j, k, grid, χxᶠᶜᶜ, c, args...) +
                                          ℑzᵃᵃᶜ(i, j, k, grid, χxᶠᶜᶜ, c, args...)
```

glwagner:
Hmm maybe there's some wrong thinking there. I'm going to go back and try to get the discrete dissipation more carefully.

glwagner:
Ok, amazingly I think what I wrote above is the correct tracer dissipation on the staggered grid. It takes a bit of algebra to figure out, and it's best to consider the one-dimensional case first. We can write this up in the docs if you like.

glwagner:
Hmm, looks like there are no docs, but we should probably change that

src/FlowDiagnostics.jl
tomchor: So far I haven't paid much attention to the conservative forms when implementing diagnostics since that sort of precision hasn't been important for the research I've done with these. But I suspect you're right!

Should we include a disclaimer in the docstrings for users until we implement the actual conservative version? (Asking for diagnostics in this package in general, not just this one.)

tomchor:
Thanks for the help @glwagner !

> Ok, amazingly I think what I wrote above is the correct tracer dissipation on the staggered grid. It takes a bit of algebra to figure out, and it's best to consider the one-dimensional case first. We can write this up in the docs if you like.

Assuming you went through the derivation to arrive at this, do you have it handy somewhere already? If so can you just paste here and then I'll transcribe it to the docs once those are set up?

> It may only return expected results for second-order operators though.

Do you mind elaborating on this comment, please?

glwagner:
> > It may only return expected results for second-order operators though.
> 
> Do you mind elaborating on this comment, please?

For higher-order operators, there are further identities (algebraic identities on the discrete staggered grid) that one can do to isolate the part of dissipation that contains no exact integrals.

For example, just multiplying the tracer equation by `c` produces a term

```
c * k * lap(c)
```

For this second-order operator, we can separate this into two terms,

```
div(k * c * grad(c)) - k * |grad(c)|^2
```

The second term is the dissipation. For higher-order operators, we need more steps.

> Assuming you went through the derivation to arrive at this, do you have it handy somewhere already?

I scribbled it on a piece of paper. Probably best to write it up when we have docs.

src/FlowDiagnostics.jl
glwagner: I think it's just that using `κᶜᶜᶜ` to calculate the tracer variance is an approximation; it isn't the exact tracer dissipation. I think for LES this could be an issue, but for DNS it may not be.

The other issue is more mundane: its actually harder to write generic code for this approximation. If we use the exact tracer variance we write pretty generic code as expressed below.

I suspect this will be true across the board --- and therein will lie a real value for this package, because it can be challenging to find the staggered-grid discrete analogs of high-order quantities that we want to calculate. So it's good to figure that out once (here).

glwagner:
A factor of two does seem large, but plausible...

It may be possible to test this. For example, consider the diffusion of a random tracer field with `advection=nothing` (so there's no numerical diffusion from the advection scheme). Then we can compare time-series of `Average(c^2/2)` to another time-series generated by integrating the variance equation with the RHS calculated from the "exact" variance we're using here (perhaps just with forward Euler to start, which should be ok if we keep the time steps very small I think).

tomchor:
Okay FYI I ran some not-so-controlled tests in one of my simulations and this new formulation seems to be farther from what I expect to be the correct value of variance dissipation, indicating it isn't correct. So we should not merge this before running some rigorous tests (like the one you mentioned).

The option you mentioned (`advection=nothing`) only works for the hydrostatic model, correct?

glwagner:
If there's no velocity field then there is no difference between the nonhydrostatic and hydrostatic model.

glwagner:
But what is the test that verifies that some other formula for the tracer variance dissipation is correct? I'm not sure I understand the reasoning.

tomchor:
Just recording some progress to come back to it later. The plotted quantities (which should be identifiable from the legend and are integrated in the domain) are from a `NonhydrostaticModel` simulation with zero velocity and random initial condition for a tracer `c`. Each panel shows two different ways of quantifying the tracer variance dissipation rate using the tools available in this PR (the top panel is offset so that both quantities start at zero). These are pretty different so something's not right:

image

Although I should note that the  ratio of both estimates is almost exactly 16, which probably means there is a factor of 2 (or more) missing somewhere...

PS: These plots ignore the molecular diffusion of tracer variance, but that term is pretty small.

glwagner:
You should be comparing c^2 / 2 right? Not c^2. That might be 1 factor of 2...

As for the other factor of 8, well thats 2^3... hm...

Can you confirm the identity

```
c * δ(q) = δ(ℑ(c) * q) - ℑ(q * δ(c))
```

?

Maybe just turn molecular diffusion off? You may also be able to use the closure tuple directly with this formulation if you have to include it.

tomchor:
> You should be comparing c^2 / 2 right? Not c^2. That might be 1 factor of 2...
> 
> As for the other factor of 8, well thats 2^3... hm...
> 
> Can you confirm the identity
> 
> ```
> c * δ(q) = δ(ℑ(c) * q) - ℑ(q * δ(c))
> ```
> 
> ?
> 
> Maybe just turn molecular diffusion off? You may also be able to use the closure tuple directly with this formulation if you have to include it.

I'll work on the identity later, but for reference, as of right now I think the comparisons are correct, no? From Stull's book:

image

The factor of 2 is included in the `TracerVarDissipRate` definition: https://github.com/tomchor/Oceanostics.jl/blob/fdad22b7deba9f12427f4405e8f14ca3c1032999/src/FlowDiagnostics.jl#L364-L369

So unless I'm missing something the correct comparison is indeed with $c^2$.

tomchor:
> Can you confirm the identity
> 
> ```
> c * δ(q) = δ(ℑ(c) * q) - ℑ(q * δ(c))
> ```
> 
> ?

Assuming `δ` is the difference operator and `ℑ` is the interpolation operator, I was able to get the same identity.

I couldn't find it online, but doesn't MITgcm have these diagnostics figured out already? Maybe we could double-check.

glwagner:
I guess it depends whether you define variance as c^2 or 1/2 * c^2. Since kinetic energy is 1/2 u^2, I've used 1/2 * c^2 by analogy. Stull also defines "epsilon" as the dissipation rate of 1/2 * c^2.

tomchor:
I spent some time today investigating this and I think we're forgetting some unit factor in the definition of the tracer variance.

To exemplify, here's a new example that uses a resolved initial noise (to minimize issues related to grid-size noise):

```julia
using Oceananigans
using Oceanostics

N = 4
κ = 1

grid = RectilinearGrid(topology=(Periodic, Periodic, Periodic),
                       size=(N,N,N), extent=(1,1,1))
model = NonhydrostaticModel(grid=grid, tracers=:c, closure=ScalarDiffusivity(κ=κ))

# A kind of convoluted way to create x-periodic, resolved initial noise
σx = 2grid.Δxᶜᵃᵃ # x length scale of the noise
σy = 2grid.Δyᵃᶜᵃ # x length scale of the noise
σz = 2grid.Δzᵃᵃᶜ # z length scale of the noise

N = 2^4 # How many Gaussians do we want sprinkled throughout the domain?
x₀ = grid.Lx * rand(N); y₀ = grid.Ly * rand(N); z₀ = -grid.Lz * rand(N) # Locations of the Gaussians

xₚ = x₀ .+ (grid.Lx .* [-2;;-1;;0;;1;;2]) # Make that noise periodic by "infinite" horizontal reflection
yₚ = y₀ .+ (grid.Ly .* [-2;;-1;;0;;1;;2]) # Make that noise periodic by "infinite" horizontal reflection
zₚ = z₀ .+ (grid.Lz .* [-2;;-1;;0;;1;;2]) # Make that noise periodic by "infinite" horizontal reflection

resolved_noise(x, y, z) = sum(@. exp(-(x-xₚ)^2/σx^2 -(y-yₚ)^2/σy^2 -(z-zₚ)^2/σz^2))
set!(model, c=resolved_noise)

simulation = Simulation(model; Δt=grid.Δxᶜᵃᵃ^2/κ/20, stop_time=0.05)

c = model.tracers.c
χ  = Oceanostics.FlowDiagnostics.IsotropicTracerVarianceDissipationRate(model, :c)
χ2 = @at (Center, Center, Center) 2 * κ * (∂x(c)^2 + ∂y(c)^2 + ∂z(c)^2) # Same as χ but non-conservative
ε_q = @at (Center, Center, Center) κ * (∂x(∂x(c^2)) + ∂y(∂y(c^2)) + ∂z(∂z(c^2)))
c² = c^2

∫χdV   = Integral(χ)
∫χ2dV  = Integral(χ2)
∫ε_qdV = Integral(ε_q)
∫c²dV  = Integral(c²)

outputs = (; χ, ∫χdV,
           χ2, ∫χ2dV,
           ε_q, ∫ε_qdV,
           c, c², ∫c²dV)

dt = 1e-4
simulation.output_writers[:tracer] = NetCDFOutputWriter(model, outputs;
                                                        filename = "tracer_diff.nc",
                                                        schedule = TimeInterval(dt),
                                                        overwrite_existing = true)
run!(simulation)

using NCDatasets, GLMakie

xc, yc, zc = nodes(c)

ds = NCDataset(simulation.output_writers[:tracer].filepath, "r")

lines(ds["time"], ds["∫χdV"], label="∫χdV (new conservative form)")
lines!(ds["time"], ds["∫χ2dV"], label="∫χ2dV (old non-conservative form)")

∂t∫c²dV = -diff(ds["∫c²dV"]) / dt
lines!(ds["time"][2:end], ∂t∫c²dV, label="∂(∫c²dV)/∂t")
axislegend()

@show ds["∫χdV"][2:end] ./ ∂t∫c²dV

close(ds)
```

With `N=16` (i.e. 16x16x16 grid), we get these results:

image

Indicating that the manual, non-conservative calculation of the tracer dissipation rate is very close to correct (based on the rate of decrease of $\int c^2 dV$), but the new calculation proposed by this PR is way off. In fact, it's a factor of almost exactly 64 off:

```julia
julia> @show mean(ds["∫χdV"][2:end] ./ ∂t∫c²dV)
mean((ds["∫χdV"])[2:end] ./ ∂t∫c²dV) = 63.703221099778375
63.703221099778375
```

When I run the same example with `N=4`, the results are similar but instead of the results being off by a factor of 64, we get a factor of 16:

```julia
julia> @show mean(ds["∫χdV"][2:end] ./ ∂t∫c²dV)
mean((ds["∫χdV"])[2:end] ./ ∂t∫c²dV) = 15.949624383944931
15.949624383944931
```

I'll sit down at some point soon and work out the units of the functions `diffusive_flux_z()`, etc., but I'm pretty sure we need to multiply each flux by its areas here:

https://github.com/tomchor/Oceanostics.jl/blob/d1983465bf8a6688f2ecc6adade895e9e191eee2/src/FlowDiagnostics.jl#L362-L367

tomchor:
Indeed after multiplying the fluxes by the appropriate area the lines all match

image

What a difference from the non-conservative form.

tomchor:
A note is that this formulation can also work for anisotropic diffusivities, but I think at the moment the `diffusive_flux_x()`  functions aren't defined for tuples since when I use a tuple closure they fall back to here and return zero: https://github.com/CliMA/Oceananigans.jl/blob/2cd91ab4f42e528981422171d6ab2f323a1ce044/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl#L158-L169

test/runtests.jl
```diff
@@ -221,8 +224,10 @@ scalar_diff = ScalarDiffusivity(ν=1e-6, κ=1e-7)
                                     tracers = :b,
                                     closure = closure)
 
-        @info "Testing energy dissipation rate terms with closure" closure
-        test_ke_dissipation_rate_terms(model)
+        if !(closure isa Tuple) || all(isa.(closure, ScalarDiffusivity{ThreeDimensionalFormulation}))
+            @info "Testing energy dissipation rate terms with closure" closure
+            test_ke_dissipation_rate_terms(model)
+        end
```
tomchor, line 230: This if clause is here because `ViscousDissipationRate` doesn't work with anisotropic closures yet, but I'm working on that

test/test_budgets.jl
```diff
@@ -0,0 +1,48 @@
...
+                                                            filename = "test_tracer_variance_budget.nc",
+                                                            schedule = IterationInterval(1),
+                                                            overwrite_existing = true)
+    run!(simulation)
+
+    ds = NCDataset(simulation.output_writers[:tracer].filepath, "r")
+    ∫∫χdVdt_final = ds["∫c²dV"][1] .- (cumsum(ds["∫χdV"]) * simulation.Δt)[end]
+    ∫c²dV_final   = ds["∫c²dV"][end]
+    close(ds)
+    @test ≈(∫∫χdVdt_final, ∫c²dV_final, rtol=rtol)
```
tomchor, line 45: I think we want to test the budget, which means integrating in time. The only way I could come up to do that is by writing the output to a file.

@glwagner is there a better way to integrate in time? Preferably without having to write anything to disk?

glwagner: You can also use `WindowedTimeAverage` and multiply by the time interval to obtain an integral

tomchor: Using `WindowedTimeAverage` would be the easiest, but it gives me this error:

```julia
julia> ∫χ̄dV = WindowedTimeAverage(∫χdV; schedule=AveragedTimeInterval(simulation.stop_time))
ERROR: MethodError: objects of type Reduction{typeof(sum!), Oceananigans.AbstractOperations.BinaryOperation{Center, Center, Center, typeof(*), KernelFunctionOperation{Center, Center, Center, NamedTuple{(:closure, :clock, :buoyancy, :id), Tuple{HorizontalScalarDiffusivity{Oceananigans.TurbulenceClosures.ExplicitTimeDiscretization, Float64, NamedTuple{(:c,), Tuple{Float64}}}, Clock{Float64}, Nothing, Val{1}}}, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64, typeof(Oceanostics.FlowDiagnostics.isotropic_tracer_variance_dissipation_rate_ccc), Tuple{Nothing, Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, NamedTuple{(:u, :v, :w, :c), Tuple{Field{Face, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}}}}}, Oceananigans.AbstractOperations.GridMetricOperation{Center, Center, Center, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64, typeof(Oceananigans.Operators.Vᶜᶜᶜ)}, typeof(Oceananigans.Operators.identity2), typeof(Oceananigans.Operators.identity3), RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64}, Tuple{Int64, Int64, Int64}} are not callable
Stacktrace:
 [1] fetch_output(output::Reduction{typeof(sum!), Oceananigans.AbstractOperations.BinaryOperation{Center, Center, Center, typeof(*), KernelFunctionOperation{Center, Center, Center, NamedTuple{(:closure, :clock, :buoyancy, :id), Tuple{HorizontalScalarDiffusivity{Oceananigans.TurbulenceClosures.ExplicitTimeDiscretization, Float64, NamedTuple{(:c,), Tuple{Float64}}}, Clock{Float64}, Nothing, Val{1}}}, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64, typeof(Oceanostics.FlowDiagnostics.isotropic_tracer_variance_dissipation_rate_ccc), Tuple{Nothing, Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, NamedTuple{(:u, :v, :w, :c), Tuple{Field{Face, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}}}}}, Oceananigans.AbstractOperations.GridMetricOperation{Center, Center, Center, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64, typeof(Oceananigans.Operators.Vᶜᶜᶜ)}, typeof(Oceananigans.Operators.identity2), typeof(Oceananigans.Operators.identity3), RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64}, Tuple{Int64, Int64, Int64}}, model::Nothing)
   @ Oceananigans.OutputWriters ~/.julia/packages/Oceananigans/IdjQp/src/OutputWriters/fetch_output.jl:10
 [2] WindowedTimeAverage(operand::Reduction{typeof(sum!), Oceananigans.AbstractOperations.BinaryOperation{Center, Center, Center, typeof(*), KernelFunctionOperation{Center, Center, Center, NamedTuple{(:closure, :clock, :buoyancy, :id), Tuple{HorizontalScalarDiffusivity{Oceananigans.TurbulenceClosures.ExplicitTimeDiscretization, Float64, NamedTuple{(:c,), Tuple{Float64}}}, Clock{Float64}, Nothing, Val{1}}}, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64, typeof(Oceanostics.FlowDiagnostics.isotropic_tracer_variance_dissipation_rate_ccc), Tuple{Nothing, Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, NamedTuple{(:u, :v, :w, :c), Tuple{Field{Face, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}}}}}, Oceananigans.AbstractOperations.GridMetricOperation{Center, Center, Center, RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64, typeof(Oceananigans.Operators.Vᶜᶜᶜ)}, typeof(Oceananigans.Operators.identity2), typeof(Oceananigans.Operators.identity3), RectilinearGrid{Float64, Periodic, Periodic, Periodic, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Float64}, Tuple{Int64, Int64, Int64}}, model::Nothing; schedule::AveragedTimeInterval, fetch_operand::Bool)
   @ Oceananigans.OutputWriters ~/.julia/packages/Oceananigans/IdjQp/src/OutputWriters/windowed_time_average.jl:178
 [3] top-level scope
   @ REPL[86]:1
 [4] top-level scope
   @ ~/.julia/packages/CUDA/BbliS/src/initialization.jl:52
```

So I think it doesn't yet work with reductions. Plus I think you can only use `WindowedTimeAverage` with a write, right? So we'd still need to write to disk.

tomchor: > define a function that integrates in time on the fly?

I think that would be the preferred method, but to be honest I don't know how to code such a function :grimacing: 

Would we use a callback for that?

glwagner: I guess you can call the `WindowedTimeAverage` functions directly to avoid a write... but you're right probably not the simplest option.

To accumulate a tracer on the fly you'd do something like

```julia
∫c = CenterField(grid)

function accumulate_c(sim)
    Δt = sim.Δt
    c = sim.model.tracers.c
    parent(∫c) .+= Δt .* parent(c)
    return nothing
end

simulation.callbacks[:integrate_c] = Callback(accumulate_c)
```

The integral `∫c` has to have the same location as the field being integrated. One could also write a general utility for this like

```julia
struct TimeIntegral{F}
    integrand :: F
    integral :: I
end

TimeIntegral(f) = TimeIntegral(f, Field(location(f)))

function (int::TimeIntegral)(sim)
    Δt = sim.Δt
    int.integral .+= Δt .* int.integrand
    return nothing
end
```

using `parent` restricts to concrete fields but I think is a bit faster, using bare broadcasting makes it work with any abstract operation.

tomchor: Thanks!

tomchor:
@glwagner I think this is ready to merge. Let me know if you have any comments/suggestions otherwise I'll merge this soon.

PR #98, opened 2022-12-01
Implements `RichardsonNumber` that respects gravity direction when domain is rotated

tomchor:
The previous version of the Richardson number was just a simple abstract operation. The new version is a `KernelFunctionOperation` that finds out what is the direction of the true gravity (through `model.buoyancy`) and calculates the richardson number accordingly. Useful for rotated domains.

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -88,7 +88,12 @@ function RichardsonNumber(model; location = (Center, Center, Face), add_backgrou
         b = model.tracers.b
     end
 
-    vertical_dir_x, vertical_dir_y, vertical_dir_z = model.buoyancy.gravity_unit_vector
+    true_vertical_direction =  model.buoyancy.gravity_unit_vector
+    if true_vertical_direction isa Oceananigans.Grids.ZDirection
+        vertical_dir_x, vertical_dir_y, vertical_dir_z = (0, 0, 1)
+    else
+        vertical_dir_x, vertical_dir_y, vertical_dir_z = true_vertical_direction
+    end
     return KernelFunctionOperation{Center, Center, Face}(richardson_number_ccf, model.grid;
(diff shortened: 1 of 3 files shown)


src/FlowDiagnostics.jl
```diff
@@ -40,15 +40,64 @@ function add_background_fields(model)
...
+    end
+
+    true_vertical_direction =  model.buoyancy.gravity_unit_vector
+    if true_vertical_direction isa Oceananigans.Grids.ZDirection
+        vertical_dir_x, vertical_dir_y, vertical_dir_z = (0, 0, 1)
+    else
+        vertical_dir_x, vertical_dir_y, vertical_dir_z = true_vertical_direction
+    end
+    return KernelFunctionOperation{Center, Center, Face}(richardson_number_ccf, model.grid;
+                                                     computed_dependencies=(u, v, w, b), parameters=(; vertical_dir_x, vertical_dir_y, vertical_dir_z))
```
glwagner, line 98: why not just pass `gravity_unit_vector` in? It's more general, you won't have to write new if statements for every new unit vector type.

tomchor: As in substitute `parameters=(; vertical_dir_x, vertical_dir_y, vertical_dir_z)` with `parameters=gravity_unity_vector`?

If so, when I tried that on the GPU I got errors, so I thought it wasn't possible. But maybe I was doing something wrong?

glwagner: It should be possible. For example, the tendency kernels in Oceananigans also use `gravity_unit_vector`.

tomchor: Hmm... I may have done something wrong when I tried this then. I'll try again since I agree it would be better.

Thanks for the tip.

tomchor: Done!

Issue #103, opened 2022-12-30
Some questions about calculating the fluctuation variables in TKE budget equation


zwei961120:
In the TKE budget equation `∂_t E = - ∂_z ⟨w′e′ + w′p′⟩ - ⟨w′u′⟩ ∂_z U - ⟨w′v′⟩ ∂_z V + ⟨w′b′⟩ - ϵ`, for example, when calculating the shear production term in x-direction, in my opinion, one need to calculate the horizontally averaged variables (U, V and W) and the fluctuation variables (u'= u-U, v'= v-V, w'= w-W) first, then calculate the shear production term [-(u’u’∂xU + v’u’∂xV + w’u’∂xW)], but in the script TKEBudgetTerms.jl, it seems like the model.velocities (u, v and w) is directly used for calculating the shear production terms:

> @inline function shear_production_x_ccc(i, j, k, grid, u, v, w, U, V, W)
>     u_int = ℑxᶜᵃᵃ(i, j, k, grid, u) # F, C, C  → C, C, C
> 
>     ∂xU = ∂xᶜᶜᶜ(i, j, k, grid, U) # F, C, C  → C, C, C
>     uu = ℑxᶜᵃᵃ(i, j, k, grid, ψ², u)
>     uu∂xU = uu * ∂xU
> 
>     ∂xV = ℑxyᶜᶜᵃ(i, j, k, grid, ∂xᶠᶠᶜ, V) # C, F, C  → F, F, C  → C, C, C
>     vu = ℑyᵃᶜᵃ(i, j, k, grid, v) * u_int
>     vu∂xV = vu * ∂xV
> 
>     ∂xW = ℑxzᶜᵃᶜ(i, j, k, grid, ∂xᶠᶜᶠ, W) # C, C, F  → F, C, F  → C, C, C
>     wu = ℑzᵃᵃᶜ(i, j, k, grid, w) * u_int
>     wu∂xW = wu * ∂xW
> 
>     return -(uu∂xU + vu∂xV + wu∂xW)
> end

Also, when calculating the pressure redistribution term, the pressure p is directly used without any operations. I don't know if my understanding is correct, and if not, are there any special considerations for doing in that way?

zwei961120:
It seems that the ZpressureRedistribution function has not been adjusted in Oceanostics.jl repo yet, right? So I tested on GPUs in the following way. The model setup is the same as the Langmuir Turbulence example setup in:
[...]

PR #104, opened 2023-01-03
Improves shear production rate calculations

tomchor:
Closes https://github.com/tomchor/Oceanostics.jl/issues/103

CC @zwei961120


tomchor:
@zwei961120 can you please check if this correctly addresses your issue?

About the pressure: I haven't ever done anything with the pressure to be honest. Do you mean that there should also be a separation between average and perturbation pressure?

src/TKEBudgetTerms.jl
```diff
@@ -268,14 +292,22 @@ end
...
 
-ZShearProduction(model; U=ZeroField(), V=ZeroField(), W=ZeroField(), kwargs...) =
-    ZShearProduction(model, model.velocities..., U, V, W; kwargs...)
+"""
+    $(SIGNATURES)
+
+Calculate the shear production rate in the `model`'s `z` direction. At least one of the mean 
+velocities `U`, `V` and `W` must be specified otherwise the output will be zero.
+"""
+function ZShearProductionRate(model; U=0, V=0, W=0, kwargs...)
```
glwagner, line 307: In principle it might be better to use `ZeroField` rather than `0` because can define `u - ZeroField() = u`. (However, I'm not sure that shortcut exists).

zwei961120:
Hi Tomas Chor,

I'm sorry for not getting back to you sooner.  I have seen the improvements you have made to the shear production rate calculation, which is the same as my ideas. Thanks!

For the pressure flux term, in my opinion, when someone wants to do TKE budget analysis in the Z direction, the pressure field needs to be averaged horizontally (P) and calculate the separation (p - P), because in the horizontally averaged TKE budget equation, the pressure flux divergence term is (-∂z(w'p')), right?  Of course, I don't know your thoughts on this point, feel free to discuss it with me.

Finally, do I need to leave comments on your improvements on GitHub, which might help more people?

Cheers,
Zheng
发件人: Tomas Chor ***@***.***>
日期: 星期二, 2023年1月3日 23:27
收件人: tomchor/Oceanostics.jl ***@***.***>
抄送: Zheng WEI ***@***.***>, Mention ***@***.***>
主题: Re: [tomchor/Oceanostics.jl] Improves on shear production rate calculations (PR #104)

@zwei961120<https://github.com/zwei961120> can you please check if this correctly addresses your issue?

About the pressure: I haven't ever done anything with the pressure to be honest. Do you mean that there should also be a separation between average and perturbation pressure?

src/TKEBudgetTerms.jl
tomchor: I agree with you that `ZeroField()` is probably better. The reason I used `0` is that currently this isn't working:

```julia
julia> model.velocities.u - Oceananigans.Fields.ZeroField()
ERROR: type ZeroField has no field grid
Stacktrace:
 [1] getproperty(x::Oceananigans.Fields.ZeroField{Int64, 3}, f::Symbol)
   @ Base ./Base.jl:38
 [2] validate_grid(a::Field{Face, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, b::Oceananigans.Fields.ZeroField{Int64, 3})
```

But I think this should be implemented/fixed on the Oceananigans side, no?

tomchor:
@zwei961120 So the code in this repo isn't merged in the main branch of Oceanostics yet. So to test this code out, the best way is to install this specific branch of the package in your Project. This can be done as:

```julia
] add Oceanostics#tc/shearprod
```

Or using 

```julia
using Pkg
Pkg.add("Oceanostics", rev="tc/shearprod")
```

tomchor:
That said, your code [here](https://github.com/zwei961120/TKE_test/blob/main/TKE_TEST.ipynb) seems to be constructing  things properly, although `wp` is never calculated, so we don't really know if it works on GPUs.

Here's a modified version of your code that (I think) would work:

```julia
using Oceananigans
using Oceananigans.Units: minute, minutes, hours
using Oceanostics
grid = RectilinearGrid(GPU();
                       size=(32, 32, 32),
                       extent=(128, 128, 64))

using Oceananigans.BuoyancyModels: g_Earth

amplitude = 0.8 # m
wavelength = 60  # m
wavenumber = 2π / wavelength # m⁻¹
frequency = sqrt(g_Earth * wavenumber) # s⁻¹

# The vertical scale over which the Stokes drift of a monochromatic surface wave
# decays away from the surface is `1/2wavenumber`, or
const vertical_scale = wavelength / 4π

# Stokes drift velocity at the surface
const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
uˢ(z) = Uˢ * exp(z / vertical_scale)
∂z_uˢ(z, t) = 1 / vertical_scale * Uˢ * exp(z / vertical_scale)
Qᵘ = -3.72e-5 # m² s⁻², surface kinematic momentum flux

u_boundary_conditions = FieldBoundaryConditions(top = FluxBoundaryCondition(Qᵘ))
Qᵇ = 2.307e-8 # m² s⁻³, surface buoyancy flux
N² = 1.936e-5 # s⁻², initial and bottom buoyancy gradient

b_boundary_conditions = FieldBoundaryConditions(top = FluxBoundaryCondition(Qᵇ),
                                                bottom = GradientBoundaryCondition(N²))

Qᵇ = 2.307e-8 # m² s⁻³, surface buoyancy flux
N² = 1.936e-5 # s⁻², initial and bottom buoyancy gradient
                                                
b_boundary_conditions = FieldBoundaryConditions(top = FluxBoundaryCondition(Qᵇ),
                        bottom = GradientBoundaryCondition(N²))

coriolis = FPlane(f=1e-4) # s⁻¹
model = NonhydrostaticModel(; grid, coriolis,
                            advection = WENO(),
                            timestepper = :RungeKutta3,
                            tracers = :b,
                            buoyancy = BuoyancyTracer(),
                            closure = AnisotropicMinimumDissipation(),
                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ),
                            boundary_conditions = (u=u_boundary_conditions, b=b_boundary_conditions))
Ξ(z) = randn() * exp(z / 4)
initial_mixed_layer_depth = 33 # m
stratification(z) = z < - initial_mixed_layer_depth ? N² * z : N² * (-initial_mixed_layer_depth)

bᵢ(x, y, z) = stratification(z) + 1e-1 * Ξ(z) * N² * model.grid.Lz

u, v, w = model.velocities
p = sum(model.pressures)
W = Field(Average(w, dims=(2, 3)))
P = Field(Average(p, dims=(2, 3)))
w_prime = w-W
p_prime = p-P

wp_operation = ZPressureRedistribution(model, w_prime, p_prime)
wp = Field(wp_operation)

compute!(wp) # This line should throw an error on GPUs if my code is too complex
```

To code, in fact, can be made much shorter since we really only need `w` and `p`. It doesn't really matter what the ICs are, etc. So a "better" (i.e. shorter) test code would be

```julia
using Oceananigans
using Oceanostics
grid = RectilinearGrid(GPU();
                       size=(32, 32, 32),
                       extent=(128, 128, 64))
model = NonhydrostaticModel(; grid)


u, v, w = model.velocities
p = sum(model.pressures)
W = Field(Average(w, dims=(2, 3)))
P = Field(Average(p, dims=(2, 3)))
w_prime = w-W
p_prime = p-P

wp_operation = ZPressureRedistribution(model, w_prime, p_prime)
wp = Field(wp_operation)

compute!(wp) # This line should throw an error on GPUs if my code is too complex
```

Note that this would only work after you install this specific branch in your project! (Like I showed you in the previous comment.)

Note also that I haven't tested the codes above. So it's possible I made a typo somewhere... :grimacing:

tomchor:
Finally, I didn't undetrstand your comment about the averaging dimensions [here](https://github.com/tomchor/Oceanostics.jl/issues/103#issuecomment-1372206756). Do you mind linking the specific part of the code you're referring to?

You make comments on specific lines of code by clicking the plus sign in [here](https://github.com/tomchor/Oceanostics.jl/pull/104/files) when you hover your mouse above it.

zwei961120:
Thanks so much for your suggestions and tips! Actually, I'm new to Julia and Github and not familiar with them, but I will keep learning😊

For the comment about the averaging dimensions, I think I’m not confused now after thinking about it, because I think how to calculate average and separation depends on specific problems and cases, thank you for your detailed reply!

tomchor:
Okay, apparently, this compiles on GPUs!:

```julia
julia> compute!(wp)
32×32×32 Field{Center, Center, Center} on RectilinearGrid on GPU
├── grid: 32×32×32 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on GPU with 3×3×3 halo
├── boundary conditions: FieldBoundaryConditions
│   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
├── operand: Derivative at (Center, Center, Center)
├── status: time=0.0
└── data: 38×38×38 OffsetArray(::CUDA.CuArray{Float64, 3, CUDA.Mem.DeviceBuffer}, -2:35, -2:35, -2:35) with eltype Float64 with indices -2:35×-2:35×-2:35
    └── max=0.0, min=0.0, mean=0.0
```

@glwagner are you okay with me merging this as is? Then when we have the `u - ZeroField() = u` shortcut implemented in Oceananigans I can change the `0`s to `ZeroField()`s.

tomchor:
> Can you summarize the changes?

Sure. The main changes are that the pressure redistribution and shear production rates calculations don't automatically use `model.velocities` for the velocities, they take user-defined velocities for the calculations. It's a necessary change because, like @zwei961120 pointed out, this means that our turbulent fluctuation velocities can now be defined accordingly to how we defined the averages. Put simply, the calculation before wasn't strictly correct.

In addition to that, I also changed the name from `ZShearProduction()` to `ZShearProductionRate()` (and same for x and y), since the term is actually the rate of the shear production, so this is a more correct name. I improved the relevant tests a bit.

glwagner:
Ah sorry can you be a little more specific? What was the interface before, and what is it now? To calculate shear production the user needs to provide both the fluctuation _and_ the mean correct?

tomchor:
> Ah sorry can you be a little more specific? What was the interface before, and what is it now? To calculate shear production the user needs to provide both the fluctuation _and_ the mean correct?

For the shear prod rate the interface used to be just

```julia
XSP = XShearProduction(model, u, v, w, U, V, W)
```

where it was up to the user to define `u`, `v`, `w`, `U`, `V` and `W` manually. Now there are two interfaces :+1: 

```julia
u, v, w = model.velocities
U = Field(Average(u, dims=(2, 3)))
V = Field(Average(v, dims=(2, 3)))
W = Field(Average(w, dims=(2, 3)))

XSP = XShearProductionRate(model, u-U, v-V, w-W, U, V, W) # manual one

XSP = XShearProductionRate(model; U=U, V=V, W=W) # this should produce the same result as before
```

As for the pressure redistribution term, the old function was this one:

```julia
function XPressureRedistribution(model)
    u, v, w = model.velocities
    p = sum(model.pressures)
    return ∂x(u*p)
```

So as you can it wasn't 100% correct unless of the averages was zero (i.e. it wasn't multiplying the turbulent fluctuation of `u` with the turbulent fluctuation of `p`; it was multiplying both full variables). The new version only has one interface that can be used as

```julia
u, v, w = model.velocities
p = sum(model.pressures)
W = Field(Average(w, dims=(2, 3)))
P = Field(Average(p, dims=(2, 3)))
w_prime = w-W
p_prime = p-P

wp = ZPressureRedistribution(model, w_prime, p_prime)
```

We could add another method for `XPressureRegistribution` where you only pass `model` and the averages.

Hope this makes sense

glwagner:
Great very clear. I think that's a good hierarchy of interfaces; defaulting to the fluctuations `model.velocities.u - U` makes sense.

glwagner:
One note is that this approximation to the shear production term is strictly only valid in the limit of vanishing time step or stationary flow. We should also check that the spatial stencil is formulated correctly to reflect the discrete conservation of kinetic energy.

glwagner:
Go for it!

src/TKEBudgetTerms.jl
tomchor: > In principle it might be better to use ZeroField rather than 0 because can define u - ZeroField() = u. (However, I'm not sure that shortcut exists).

@glwagner Do you have any reference that I can read about implementing such shortcut? I don't really know how to get started but I'd like to open a PR on Oceananigans to add that.

PR #106, opened 2023-01-03
Run CI only on commits

tomchor:
This PR removes the directive to run CI on PRs. 
I'm doing this because the same CI is run twice every time I commit (since a PR is always associated with a commit). Now it only runs once per commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: CI
 on:
   - push
-  - pull_request
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


navidcy (approved):
we have it bit different in

https://github.com/CliMA/ParameterEstimocean.jl/blob/main/.github/workflows/ci.yml

but I think what you suggest is equivalent!

tomchor:
> we have it bit different in
> 
> https://github.com/CliMA/ParameterEstimocean.jl/blob/main/.github/workflows/ci.yml
> 
> but I think what you suggest is equivalent!

Interesting! Let's try this version first and see how it goes. If there are any unintended effects we can try the ParameterEstimocean version

PR #110, opened 2023-03-13
Adds conservative form of the KE dissipation rate

tomchor:


test/test_kedissip.jl
glwagner, line 1: `test_ke_dissipation_rate.jl`?

tomchor: This file is there for now just to make my life easier, but it'll be merged with `test_budgets.jl` as a function once it's ready!

(sorry, maybe I should have marked the PR as a draft)

tomchor:
Leaving this for reference for future self: running this for longer still produces results that are kinda far from the correct ones, indicating that the formulation still isn't correct:

image

Furthermore, this is a diffusivity-dominated regime. It would be nice to test this in an advection-dominated one.

tomchor:
I improved the time-integration and results are now better. The in figure below, top panel, I'm comparing the KE, the new (supposedly conservative) formulation, the old (non-conservative formulation) of the time-integrated KE dissipation rate, and the whole diffusive term directly from `Oceananigans.Models.NonhydrostaticModel`. The bottom panel shows time-derivatives of the same quantities:

image

The dissipation rate estimates all match up really well with each other. Given that one of those rates comes from straight from `Oceananigans.Models.NonhydrostaticModel` I take them to be correct (at least for this example). It's a bit surprising that they match _that_ well with the measured KE. Is that the limit of predictability due to the interpolations?

tomchor:
`KineticEnergyDissipationRate` is now also implemented in conservative form. Tested with `NonhydrostaticModel` up to `eps()` :tada: 

I also renamed `ViscousDissipationRate` to `KineticEnergyDissipationRate`, which is more informative.

PR #118, opened 2023-04-11
Add `TracerVarianceBudgetTerms` module

tomchor:
This does two things:

1. Adds `TracerVarianceBudgetTerms` module (along with `TracerVarianceTendency` and `TracerVarianceDiffusiveTerm`)
2. Simplifies organization of tests, while making them broader

Point 1 creates a module dedicated for tracer variance budgets and uses Oceananigans kernels directly to calculate the terms. I'll add more terms slowly one by one, but using Oceananigans kernels should make this process easy and (since there are no interpolations to get tracer variance in `(Center, Center, Center)`, we can guarantee that the terms are conserve tracer variance up to machine precision. 

In fact, one advantage is that this gives a robust point of comparison for our tracer variance dissipation rate, whose average is now tested against `TracerVarianceTendency` and `TracerVarianceDiffusiveTerm` up to machine precision to ensure that it's formulation is conservative.

Point 2 makes the testing code much simpler, and now we're also going to test everything for every combination between several closures, model types, and grid types. Something like this:

```julia
@testset "Oceanostics" begin
    for grid in (regular_grid, stretched_grid)
        for model_type in (NonhydrostaticModel, HydrostaticFreeSurfaceModel)
            for closure in closures
                #Test things
...
```

This is likely overkill, but it does make the code much more readable and the code case for now is small enough that tests still run pretty fast. (Something like 20 minutes on my laptop, not counting the budget tests, which have to evolve a simulation.)

### DIFF
--- a/src/TracerVarianceBudgetTerms.jl
+++ b/src/TracerVarianceBudgetTerms.jl
@@ -14,9 +14,7 @@ using Oceananigans.TurbulenceClosures: ∇_dot_qᶜ
 
 import Oceananigans.TurbulenceClosures: diffusive_flux_x, diffusive_flux_y, diffusive_flux_z
 
-
-
-#+++ Tracer vriance tendency
+#+++ Tracer variance tendency
 @inline function c∂ₜcᶜᶜᶜ(i, j, k, grid, val_tracer_index::Val{tracer_index},
                                         val_tracer_name,
                                         advection,
@@ -42,6 +40,23 @@ import Oceananigans.TurbulenceClosures: diffusive_flux_x, diffusive_flux_y, diff
                                                                 args...)
 end
 
+"""
+    $(SIGNATURES)
+
+Return a `KernelFunctionOperation` that computes the tracer variance tendency:
+
+    TEND = 2 c ∂ₜc
+
+where `c` is the tracer and `∂ₜc` is the tracer tendency (computed using
+Oceananigans' tracer tendency kernel).
+
+```julia
+grid = RectilinearGrid(size=(4, 4, 4), extent=(1, 1, 1))
+model = NonhydrostaticModel(grid=grid, tracers=:b, closure=SmagorinskyLilly())
+
+DIFF = TracerVarianceTendency(model, :b)
+```
+"""
 function TracerVarianceTendency(model::NonhydrostaticModel, tracer_name; location = (Center, Center, Center))
     validate_location(location, "TracerVarianceTendency")
     tracer_index = findfirst(n -> n === tracer_name, propertynames(model.tracers))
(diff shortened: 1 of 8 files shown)


tomchor:
For future reference and clarity:

In this PR I'm also making the requirements for the budget test (where we evolve a simulation, time integrate the tracer variance dissipation and compare it with the tracer variance) more relaxed. This follows a discussion on slack about making "top-down" estimates (in this case the tracer variance computed as `c^2`) match with "bottom-up" estimates (i.e. the time-integrated tracer variance dissipation rate).

The main takeaway from that discussion for me was that, unless we get the calculation of (i) the tracer variance dissipation rate and (ii) the time-integration scheme to exactly match what Oceananigans does, the results will depend on grid resolution and time-step.

The previous budget test was trying to get both (i) and (ii) right, and failing to do so exactly (particularly is was failing on (ii), which is far from trivial). However, (ii) is not a goal of this package (instead that is up to the users) but (i) is. So we are now testing (i) up to machine precision by intercomparing results from one snapshot directly with Oceananigans kernels, which is a much better (_and_ faster) way to do things imo.

I'm not 100% sure what to do with budget evolution, but for now I'm increasing the tolerance of the test to 2% and adjusting the time-step with a `TimeStepWizard` which makes it run faster and still makes sure that the top-down estimates _approximately_ match the bottom-up ones. I think that's good compromise, although I'm not not totally against removing that test altogether (especially if tests start to become slow in the future).

glwagner:
This paper might be relevant to calculations of numerical + explicit dissipation:

https://www.sciencedirect.com/science/article/pii/S1463500314000754?casa_token=WpWBDmsmS-EAAAAA:BMPNh2q4glKqfTbmcs_1qnR1EB_rmv7e96cxWIwiTVNyHVOeQpTB7aznm2wKOouidG_a6YFLYw

I believe this is essentially an unsolved problem though a number of people are working on it.

src/TracerVarianceBudgetTerms.jl
```diff
@@ -0,0 +1,201 @@
...
+    return 2 * tracer[i,j,k] * ∇_dot_qᶜ(i, j, k, grid, closure, diffusivities, val_tracer_index, tracer, args...)
+end
+
+"""
+    $(SIGNATURES)
+
+Return a `KernelFunctionOperation` that computes the diffusive term of the tracer variance
+prognostic equation using Oceananigans' diffusive tracer flux divergence kernel:
+
+    DIFF = c ∂ⱼFⱼ
```
tomchor, line 98:
```suggestion
    DIFF = 2 c ∂ⱼFⱼ
```

PR #125, opened 2023-05-09
Adds Q (from Q-criterion) calculation

tomchor:
This PR adds the calculation of the "Q" velocity gradient tensor invariant. This is useful for the identification and visualization of vortices in fluid dynamics (a comparative study of similar methods can be found [here](https://www.tandfonline.com/doi/pdf/10.1080/19942060.2020.1816496#:~:text=The%20Q%20criterion%20defines%20Q,strain%20in%20the%20vortices%20areas.&text=Lambda%2D2%20criterion%20(Jeong%20%26,%2B%202%20with%20%CF%832%20%3C%200).).)

The definition and nomenclature comes from the equation for the eigenvalues of the velocity gradient tensor $\nabla \vec u$:

```math
\lambda^3+P \lambda^2+Q \lambda+T=0
```

$Q$ is then defined as

```math
Q=\frac{1}{2}\left(\|\bf\Omega\|^2-\|\bf S\|^2\right)
```

where $\bf S$ is the strain rate tensor and $\bf \Omega$ is the vorticity tensor (i.e. the symmetric and antisymmetric parts of the velocity gradient tensor, respectively).

This PR adds `KernelFunctionOperations` for `StrainRateTensorModulus` ($|| \bf S ||$), `VorticityTensorModulus` ($|| \bf \Omega ||$),  and `Q`.

PS: I don't love this nomenclature since it gives basically zero information on what it's called, but I haven't ever seen it being called something else. If anyone knows of a better name for this, lmk.

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -334,6 +334,15 @@ end
 """
     $(SIGNATURES)
 
+Calculate the modulus (absolute value) of the strain rate tensor `S`, which is defined as the
+symmetric part of the velocity gradient tensor:
+
+    Sᵢⱼ = ½(∂ⱼuᵢ + ∂ᵢuⱼ)
+
+Its modulus is then defined (using Einstein summation notation) as
+
+    || Sᵢⱼ || = √( Sᵢⱼ Sᵢⱼ)
+
 """
 function StrainRateTensorModulus(model; location = (Center, Center, Center))
     validate_location(location, "StrainRateTensorModulus", (Center, Center, Center))
@@ -358,6 +367,15 @@ end
 """
     $(SIGNATURES)
 
+Calculate the modulus (absolute value) of the vorticity tensor `Ω`, which is defined as the
+antisymmetric part of the velocity gradient tensor:
+
+    Ωᵢⱼ = ½(∂ⱼuᵢ - ∂ᵢuⱼ)
+
+Its modulus is then defined (using Einstein summation notation) as
+
+    || Ωᵢⱼ || = √( Ωᵢⱼ Ωᵢⱼ)
+
 """
 function VorticityTensorModulus(model; location = (Center, Center, Center))
     validate_location(location, "VorticityTensorModulus", (Center, Center, Center))
@@ -375,6 +393,20 @@ end
 """
     $(SIGNATURES)
 
+Calculate the value of the `Q` velocity gradient tensor invariant. This is usually just called `Q`
+and it is generally used for identifying and visualizing vortices in fluid flow.
+
+The definition and nomenclature comes from the equation for the eigenvalues `λ` of the velocity
+gradient tensor `∂ⱼuᵢ`:
+
+    λ³ + P λ² + Q λ + T = 0
+
+from where `Q` is defined as
+
+    Q = ½ ( ΩᵢⱼΩᵢⱼ - SᵢⱼSᵢⱼ)
+
+and where `Sᵢⱼ= ½(∂ⱼuᵢ + ∂ᵢuⱼ)` and `Ωᵢⱼ= ½(∂ⱼuᵢ - ∂ᵢuⱼ)`. More info about it can be found in
+10.1063/1.5124245.
 """
 function QVelocityGradientTensorInvariant(model; location = (Center, Center, Center))
     validate_location(location, "QVelocityGradientTensorInvariant", (Center, Center, Center))
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -317,7 +317,7 @@ end
 #---
 
 #+++ Velocity gradient tensor
-@inline fψ_plus_gφ²(i, j, k, grid, f, ψ, g, φ) = @inbounds (f(i, j, k, grid, ψ) + g(i, j, k, grid, φ))^2
+@inline fψ_plus_gφ²(i, j, k, grid, f, ψ, g, φ) = (f(i, j, k, grid, ψ) + g(i, j, k, grid, φ))^2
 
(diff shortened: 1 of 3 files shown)


src/FlowDiagnostics.jl
```diff
@@ -315,4 +316,104 @@ function DirectionalErtelPotentialVorticity(model, direction; location = (Face,
...
+
+Its modulus is then defined (using Einstein summation notation) as
+
+    || Sᵢⱼ || = √( Sᵢⱼ Sᵢⱼ)
+
+"""
+function StrainRateTensorModulus(model; location = (Center, Center, Center))
+    validate_location(location, "StrainRateTensorModulus", (Center, Center, Center))
+    return KernelFunctionOperation{Center, Center, Center}(strain_rate_tensor_modulus_ccc, model.grid, model.velocities...)
+end
```
navidcy, line 350: seems like kwarg `location` is not used?

src/FlowDiagnostics.jl
```diff
@@ -315,4 +316,104 @@ function DirectionalErtelPotentialVorticity(model, direction; location = (Face,
...
+
+
+@inline fψ_minus_gφ²(i, j, k, grid, f, ψ, g, φ) = @inbounds (f(i, j, k, grid, ψ) - g(i, j, k, grid, φ))^2
```
navidcy, line 353:
```suggestion
@inline fψ_minus_gφ²(i, j, k, grid, f, ψ, g, φ) = (f(i, j, k, grid, ψ) - g(i, j, k, grid, φ))^2
```

src/FlowDiagnostics.jl
```diff
@@ -315,4 +316,104 @@ function DirectionalErtelPotentialVorticity(model, direction; location = (Face,
 end
 #---
 
+#+++ Velocity gradient tensor
+@inline fψ_plus_gφ²(i, j, k, grid, f, ψ, g, φ) = @inbounds (f(i, j, k, grid, ψ) + g(i, j, k, grid, φ))^2
```
navidcy, line 320:
```suggestion
@inline fψ_plus_gφ²(i, j, k, grid, f, ψ, g, φ) = (f(i, j, k, grid, ψ) + g(i, j, k, grid, φ))^2
```

src/FlowDiagnostics.jl
```diff
@@ -315,4 +316,104 @@ function DirectionalErtelPotentialVorticity(model, direction; location = (Face,
...
+
+Its modulus is then defined (using Einstein summation notation) as
+
+    || Ωᵢⱼ || = √( Ωᵢⱼ Ωᵢⱼ)
+
+"""
+function VorticityTensorModulus(model; location = (Center, Center, Center))
+    validate_location(location, "VorticityTensorModulus", (Center, Center, Center))
+    return KernelFunctionOperation{Center, Center, Center}(vorticity_tensor_modulus_ccc, model.grid, model.velocities...)
+end
```
navidcy, line 383: similarly here, `location` kwarg is only validated but never used?

src/FlowDiagnostics.jl
```diff
@@ -315,4 +316,104 @@ function DirectionalErtelPotentialVorticity(model, direction; location = (Face,
...
+
+    Q = ½ ( ΩᵢⱼΩᵢⱼ - SᵢⱼSᵢⱼ)
+
+and where `Sᵢⱼ= ½(∂ⱼuᵢ + ∂ᵢuⱼ)` and `Ωᵢⱼ= ½(∂ⱼuᵢ - ∂ᵢuⱼ)`. More info about it can be found in
+10.1063/1.5124245.
+"""
+function QVelocityGradientTensorInvariant(model; location = (Center, Center, Center))
+    validate_location(location, "QVelocityGradientTensorInvariant", (Center, Center, Center))
+    return KernelFunctionOperation{Center, Center, Center}(Q_velocity_gradient_tensor_invariant_ccc, model.grid, model.velocities...)
+end
```
navidcy, line 414: same remark about `location`

navidcy (approved):

src/FlowDiagnostics.jl
tomchor: Just so that I understand, the reasoning here is that since there isn't a call to `getindex` here `@inbounds` is unnecessary?

navidcy: Yes, `i, j, k` are just args in a function. There is no `getindex` involved.

src/FlowDiagnostics.jl
tomchor: Yes. The idea throughout these functions is to allow for a `location` argument. At the moment, however, since these have to implemented by hand using `Oceananigans.Operators` (and this being a new package with contributions mainly by myself), each diagnostic is only implemented in one location (most of them are at cell centers, although at least one (PV) is at cell edges). Hopefully, as things are needed and perhaps contributions come, there will be more location options for diagnostics.

So, in summary, that kwarg is there in anticipation of a future when we will have more options. One could certainly make the case to remove it, while adding superscripts in Oceananigans' style that indicates the location. So `VorticityTensorModulus` would become `VorticityTensorModulusᶜᶜᶜ` and the `location` wouldn't exist, but then I think the user interface suffers a bit.

If you have suggestions about other ways to manage this lmk! I'm currently not happy with either solution tbh

navidcy: So you are saying that the problem is that `vorticity_tensor_modulus_ccc` needs to be redefine for each location, right?

tomchor: Yes. `vorticity_tensor_modulus_ccf`, `vorticity_tensor_modulus_cfc` and so on...

PR #127, opened 2023-05-15
Fix typo + minor suggestions in `README`

navidcy:


tomchor (approved):
thanks!

(Although these lines may be overwritten by the docs PR; not sure)

PR #128, opened 2023-05-18
Some code cleanup + add a missing `@inbounds`

navidcy:
Most changes are aesthetic and less code.

Only adding a few `@inbounds` is possibly the important changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.9'
+          - '1.8'
         os:
           - ubuntu-latest
         arch:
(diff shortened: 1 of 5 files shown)


tomchor:
Thanks! I wonder how much difference the `@inbounds` makes...

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.8'
+          - '1.9'
```
tomchor, line 12: I'm reluctant to test a Julia version that Oceananigans isn't tested against yet, and not test the version that Oceananigans _is_ tested agaisnt.

navidcy: Damn! That was a mistake! I was meant to do that for a different repo and did it here apparently... I'll remove this! :)

tomchor (approved):
Thanks! This looks great

PR #132, opened 2023-05-20
Ensure that output files are deleted before pushing to `gh-pages`

navidcy:
At the moment there are some `.nc` files pushed up there and in the course of time this can make the repo grow in size a lot.


tomchor (approved):
Thanks!

navidcy:
We can generalize (e.g., like in https://github.com/CliMA/Oceananigans.jl/pull/3113) but for now I'm merging and we see ;)

PR #136, opened 2023-06-07
Adds `KineticEnergyForcingTerm`

tomchor:
This PR adds `KineticEnergyForcingTerm`, which calculates the contribution of the momentum forcing terms on the kinetic energy evolution. It's especially useful when calculating the dissipation rate due to sponge layers and such.

### DIFF
--- a/src/TKEBudgetTerms.jl
+++ b/src/TKEBudgetTerms.jl
@@ -2,7 +2,7 @@ module TKEBudgetTerms
 using DocStringExtensions
 
 export TurbulentKineticEnergy, KineticEnergy
-export KineticEnergyTendency, KineticEnergyDiffusiveTerm
+export KineticEnergyTendency, KineticEnergyDiffusiveTerm, KineticEnergyForcingTerm
 export IsotropicKineticEnergyDissipationRate, KineticEnergyDissipationRate
 export XPressureRedistribution, YPressureRedistribution, ZPressureRedistribution
 export XShearProductionRate, YShearProductionRate, ZShearProductionRate
@@ -25,8 +25,8 @@ using Oceanostics: validate_location, validate_dissipative_closure
 # Some useful operators
 @inline ψ²(i, j, k, grid, ψ) = @inbounds ψ[i, j, k]^2
 
-@inline ψ′²(i, j, k, grid, ψ, Ψ) = @inbounds (ψ[i, j, k] - Ψ[i, j, k])^2
-@inline ψ′²(i, j, k, grid, ψ, Ψ::Number) = @inbounds (ψ[i, j, k] - Ψ)^2
+@inline ψ′²(i, j, k, grid, ψ, ψ̄) = @inbounds (ψ[i, j, k] - ψ̄[i, j, k])^2
+@inline ψ′²(i, j, k, grid, ψ, ψ̄::Number) = @inbounds (ψ[i, j, k] - ψ̄)^2
 
 @inline fψ²(i, j, k, grid, f, ψ) = f(i, j, k, grid, ψ)^2
 
@@ -260,7 +260,7 @@ end
 """
     $(SIGNATURES)
 
-Return a `KernelFunctionOperation` that computes the diffusive term the KE prognostic equation:
+Return a `KernelFunctionOperation` that computes the diffusive term of the KE prognostic equation:
 
 ```
     DIFF = uᵢ∂ⱼτᵢⱼ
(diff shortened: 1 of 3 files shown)


src/TKEBudgetTerms.jl
```diff
@@ -285,6 +285,44 @@ function KineticEnergyDiffusiveTerm(model; location = (Center, Center, Center))
...
+where `uᵢ` are the velocity components and `Fᵤᵢ` is the forcing term(s) in the `uᵢ`
+prognostic equation (i.e. the forcing for `uᵢ`).
+"""
+function KineticEnergyForcingTerm(model::NonhydrostaticModel; location = (Center, Center, Center))
+    validate_location(location, "KineticEnergyForcingTerm")
+    model_fields = fields(model)
+
+    if model isa HydrostaticFreeSurfaceModel
+        model_fields = (; model_fields..., w=ZeroField())
+    end
```
glwagner, line 318: If there isn't a function that generalizes for different models, we should add it in Oceananigans.

glwagner:
Maybe `ForcingProduction` term?

tomchor:
> Maybe `ForcingProduction`? (Are we adding the word "Term" to the end of all the names?)

I haven't been very good about following any patterns, which basically means I add the ending `Term` when I notice that people generally tend to refer to that term as a _term_.

For example people usually say "I'll calculate the KE dissipation rate", not "I'll calculate the KE dissipation rate _term_", so that term is called only `KineticEnergyDissipationRate`, as opposed to `KineticEnergyDissipationRateTerm`.

For the diffusive term, I feel like people usually do refer to that as the "diffusive _term_", so its name is `KineticEnergyDiffusiveTerm`. I think the same goes for the forcing production term, which is why `Term` is there.

If you have a general naming pattern suggestion, I'd be interested in hearing it! If we agree then we can make it official and start following it.

With that said, I've been going for clarity at the expense of being verbose, so I'd prefer naming it `KineticEnergyForcingProduction` (or that + `Term`) than just `ForcingProduction`.

glwagner:
> > Maybe `ForcingProduction`? (Are we adding the word "Term" to the end of all the names?)
> 
> I haven't been very good about following any patterns, which basically means I add the ending `Term` when I notice that people generally tend to refer to that term as a _term_.

Aren't all of these objects "terms"? "Term" means that the object appears among other similar ones in an equation (eg an equation consists of "terms").

glwagner:
I think we use "term" when its in the context of an equation. Ie, the TKE budget has three terms: production, buoyancy flux, dissipation rate. We might then say, "the production term does X..."

Outside the context of a budget or an equation we would not use the word "term". Personally I think we should remove the word term. "KineticEnergyDiffusiveTerm" -- is that "KineticEnergyTransport"?

tomchor:
> I think we use "term" when its in the context of an equation. Ie, the TKE budget has three terms: production, buoyancy flux, dissipation rate. We might then say, "the production term does X...", or "the dissipation rate term does Y..."

Good point.

> Outside the context of a budget or an equation we would not use the word "term". Personally I think we should remove the word term. "KineticEnergyDiffusiveTerm" -- is that "KineticEnergyTransport"?

The diffusive term includes both KE dissipation and transport, so I don't think calling it `KineticEnergyTransport` would be correct.

tomchor:
@glwagner I like this discussion. I'm tempted to merge this as is and maybe we can discuss this on an issue? I think several diagnostics need to be changed once we agree on a naming pattern, so it's good to debate that on a separate issue and implement it on a separate PR.

test/runtests.jl
```diff
@@ -205,6 +205,30 @@ function test_ke_dissipation_rate_terms(grid; model_type=NonhydrostaticModel, cl
...
+    set!(model, u=grid_noise, v=grid_noise, w=grid_noise)
+
+    ε = KineticEnergyForcingTerm(model)
+    @compute ε_field = Field(ε)
+    @test ε isa AbstractOperation
+    @test ε_field isa Field
+
+    @compute ε_truth = Field(@at (Center, Center, Center) (-0.1 * model.velocities.u^2 -0.2 * model.velocities.v^2 -0.3 * model.velocities.w^2))
+
+    @test ≈(Array(interior(ε_field, 1, 1, 1)), Array(interior(ε_truth, 1, 1, 1)), rtol=1e-12, atol=eps())
```
glwagner, line 227:
```suggestion
    @test isapprox(Array(interior(ε_field, 1, 1, 1)), Array(interior(ε_truth, 1, 1, 1)), rtol=1e-12, atol=eps())
```

glwagner (approved):

PR #141, opened 2023-06-20
Trying out improvements to docs/make.jl

tomchor:
This PR tries out a couple of changes in the `docs/make.jl` file:

1. Adds the main project's environment to the docs (this prevents the versions of deps in `Manifest.toml` and `docs/Manifest.toml` from being different, which was happening for Oceananigans)
2. Pushes previews to a different branch, so that hopefully we can still see previews, but cleaning them up will be a matter of deleting the branch

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -3,8 +3,4 @@ CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Glob = "c27321d9-0574-5035-807b-f59d2c89b15c"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
-Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 Rasters = "a3a2b9e3-a471-40c9-b274-f788e487c689"
-
-[compat]
-Oceananigans = "0.80, 0.81, 0.82"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -67,7 +67,7 @@ if CI
                versions = ["stable" => "v^", "v#.#.#", "dev" => "dev"],
                devbranch = "main",
                push_preview = true,
-               repo_previews = "github.com/tomchor/Oceanostics.jl.git:doc-previews",
+               branch_previews = "doc-previews",
                )
 end
 #---
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,3 +1,5 @@
+pushfirst!(LOAD_PATH, joinpath(@__DIR__, "..")) # add Oceananigans to environment stack
+
 using Documenter
 using Literate
 using Glob
@@ -65,6 +67,7 @@ if CI
                versions = ["stable" => "v^", "v#.#.#", "dev" => "dev"],
                devbranch = "main",
                push_preview = true,
+               repo_previews = "doc-previews",
                )
 end
 #---
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -56,7 +56,22 @@ makedocs(sitename = "Oceanostics.jl",
 #---
 
 #+++ Cleanup any output files, e.g., .jld2 or .nc, created by docs. Otherwise they are pushed up in the docs branch in the repo
-for file in vcat(glob("docs/*.jld2"), glob("docs/*.nc"), glob("docs/generated/*.nc"), glob("docs/generated/*.nc"))
+"""
+    recursive_find(directory, pattern)
+
+Return list of filepaths within `directory` that contains the `pattern::Regex`.
+"""
+recursive_find(directory, pattern) =
+    mapreduce(vcat, walkdir(directory)) do (root, dirs, files)
+        joinpath.(root, filter(contains(pattern), files))
+    end
+
+files = []
+for pattern in [r"\.jld2", r"\.nc"]
+    global files = vcat(files, recursive_find(@__DIR__, pattern))
+end
+
+for file in files
     rm(file)
 end
 #---
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -67,7 +67,7 @@ if CI
                versions = ["stable" => "v^", "v#.#.#", "dev" => "dev"],
                devbranch = "main",
                push_preview = true,
-               repo_previews = "doc-previews",
+               repo_previews = "github.com/tomchor/Oceanostics.jl.git:doc-previews",
                )
 end
 #---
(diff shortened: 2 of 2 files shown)


tomchor:
Pushing the previews to the `doc-previews` branch in the sense that the branch gets created and the [docs do appear there](https://github.com/tomchor/Oceanostics.jl/tree/doc-previews). But, as expected, the preview doesn't appear in https://tomchor.github.io/Oceanostics.jl/previews/PR141 since github pages is served by the `gh-pages` branch.

However, you can still check the branch out locally and inspect the docs, which I think is better than the alternatives (i.e. not having a preview at all, or pushing previews to the `gh-pages` branch).

tomchor:
@navidcy I also realized that the prevoius `make.jl` wasn't actually deleting the .nc and .jld2 files. (I fixed this in https://github.com/tomchor/Oceanostics.jl/pull/141/commits/5d122486b6737020697d959028c765d46ecb8e11 I think using your algorithm!)

As a result the repo size is already pretty big (I think ~200 MB?). I think you cleaned the OceananigansDocumentation repo before under similar circumstances, no? Can you point me to a simple guide on how to do this cleaning? Every how-to I see online ends up being more complex than I'm comfortable with (i.e. I don't know enough git to be sure I won't mess up important things :grimacing:).

docs/Project.toml
```diff
@@ -3,8 +3,4 @@ CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Glob = "c27321d9-0574-5035-807b-f59d2c89b15c"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
-Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 Rasters = "a3a2b9e3-a471-40c9-b274-f788e487c689"
-
-[compat]
-Oceananigans = "0.80, 0.81, 0.82"
```
navidcy, line 10: why?

navidcy: I'd say this should be the same as
https://github.com/tomchor/Oceanostics.jl/blob/5d122486b6737020697d959028c765d46ecb8e11/Project.toml#L13C10-L13C10
?

navidcy:
I don't see the `docs/Manifest.toml`.

navidcy:
Are all the large files you'd like to get rid off in a branch _other_ than `main`?

navidcy:
> @navidcy I also realized that the prevoius `make.jl` wasn't actually deleting the .nc and .jld2 files. (I fixed this in [5d12248](https://github.com/tomchor/Oceanostics.jl/commit/5d122486b6737020697d959028c765d46ecb8e11) I think using your algorithm!)
> 
> As a result the repo size is already pretty big (I think ~200 MB?). I think you cleaned the OceananigansDocumentation repo before under similar circumstances, no? Can you point me to a simple guide on how to do this cleaning? Every how-to I see online ends up being more complex than I'm comfortable with (i.e. I don't know enough git to be sure I won't mess up important things 😬).

Are all the large files you'd like to get rid off in a branch _other_ than `main`?

docs/Project.toml
tomchor: That's exactly the effect it does, along with the addition of the line 

```
pushfirst!(LOAD_PATH, joinpath(@__DIR__, "..")) # add Oceanostics environment
```

The only difference (I think) is that I don't need to write it twice. It automatically pulls it from `Project.toml`.

tomchor:
> I don't see the `docs/Manifest.toml`.

There isn't one. It's only created when the docs run (but it's never committed). I think this was something you suggested, but we can commit a `docs/Manifest.toml` if you think it's best.

tomchor:
> > @navidcy I also realized that the prevoius `make.jl` wasn't actually deleting the .nc and .jld2 files. (I fixed this in [5d12248](https://github.com/tomchor/Oceanostics.jl/commit/5d122486b6737020697d959028c765d46ecb8e11) I think using your algorithm!)
> > As a result the repo size is already pretty big (I think ~200 MB?). I think you cleaned the OceananigansDocumentation repo before under similar circumstances, no? Can you point me to a simple guide on how to do this cleaning? Every how-to I see online ends up being more complex than I'm comfortable with (i.e. I don't know enough git to be sure I won't mess up important things grimacing).
> 
> Are all the large files you'd like to get rid off in a branch _other_ than `main`?

Unless I'm missing something, yes. They're all the output files that are in `gh-pages`.

navidcy:
> Unless I'm missing something, yes. They're all the output files that are in `gh-pages`.

OK, we can deal with this easily. The tricky git bits come when you want to delete a file but keep all the rest of the git history. That involves modifying every git commit! But with `gh-pages` nobody cares about the history of the commits. We can create a new orphan branch and then copy all the contents from `gh-pages` *leaving out the big files*, commit and push. I've done this many times --  this is how I'm cleaning the `previews` history from the `gh-pages` branch in OceananigansDocumentation repo.

docs/Project.toml
navidcy: Gotcha!

Are you sure pushing the main project environment has the desired effect even docs call `Pkg.develop` later?

https://github.com/tomchor/Oceanostics.jl/blob/5d122486b6737020697d959028c765d46ecb8e11/.github/workflows/documentation.yml#L25

tomchor: According to my local test (i.e. pretty much just doing locally the same commands that github actions does), it appears to work as intended. The tests are also passing, so I'm assuming it's okay. Is there anything else I gotta test?

tomchor:
> > Unless I'm missing something, yes. They're all the output files that are in `gh-pages`.
> 
> OK, we can deal with this easily. The tricky git bits come when you want to delete a file but keep all the rest of the git history. That involves modifying every git commit! That would be the case if we had accidentally committed a `.nc` file on `main`.
> 
> But with `gh-pages` nobody cares about the history of the commits. We can create a new orphan branch and then copy all the contents from `gh-pages` _leaving out the big files_, commit and push. I've done this many times -- this is how I'm cleaning the `previews` history from the `gh-pages` branch in OceananigansDocumentation repo.

Okay, I'm not sure if I follow everything, but here's my best guess.

1. Locally create an orphan branch with `git checkout --orphan gh-pages`
2. Copy the contents of `gh-pages` without the big files over to it
3. Delete `gh-pages` from github (obviously keep it locally)
4. Push the local orphan branch to github's `gh-pages`, so replace the old one.

Is that what you mean?

Also, if it's easier for you to just do it than it is to explain to me, feel free! I think all the files are `.nc` in the `generated` directory.

e.g.: https://github.com/tomchor/Oceanostics.jl/tree/gh-pages/v0.11.2/generated

navidcy:
I can do it!

It's not quite that... I would be a bit more cautious... I would create an orphan branch called `gh-pages-new` and push that up and when all is good rename the branches to `gh-pages` -> `gh-pages-old` and `gh-pages-new` -> `gh-pages`. Then ensure the GitHub pages deploys the new branch and only after delete the old one.

navidcy:
I'll do it after this PR is merged ;)

navidcy (approved):

PR #164, opened 2024-03-22
Add `PotentialEnergy` `KernelFunctionOperation`

### LINKED ISSUE #163: Potential energy `KernelFunctionOperation`
Thanks for this great package! This is more a discussion/question than an issue.

With Oceananigans.jl v0.89 a [`seawater_density`](https://clima.github.io/OceananigansDocumentation/dev/appendix/library/#Oceananigans.Models.seawater_density-NTuple{5,%20Any}) function was added.
In my own work, one way I used this was to calculate the gravitational potential energy as a `KernelFunctionOperation` during a simulation

```julia
using Oceananigans: Models.seawater_density

@inline function potential_energy(model)

    grid = model.grid
    ρ = seawater_density(model)
    Z = Oceananigans.Models.model_geopotential_height(model)
    g = tuple(model.buoyancy.model.gravitational_acceleration)

    return KernelFunctionOperation{Center, Center, Center}(Ep, grid, ρ, Z, g)
end
@inline Ep(i, j, k, grid, ρ, Z, g) = g[1] * ρ[i, j, k] * Z[i, j, k]
```

The volume integral of this can then be saved during a `simulation`

```julia
Eₚ = potential_energy(model)
∫Eₚ = Integral(Eₚ)
```

This is a bit rough around the edges but if it is something that you think would fit in to this package let me know and I can tidy it up and add some tests. No problem at all if you think this package is not the place for it though!

My goal with this was to calculate the LHS terms in the energy budget for a closed system using equations (21), (22), (23) from [Winters et al. (1995)](https://doi.org/10.1017/S002211209500125X), your `KineticEnergy` function and the above potential energy function. One thing I could not quite get to work was calculating a background state for the potential energy (equation (22)) which depends on a resorting and scalar indexing on GPU's. Being able to calculate the energy budget like in [Winters et al. (1995)](https://doi.org/10.1017/S002211209500125X) would be a neat functionality though!

jbisits:
This PR adds the `KernelFunctionOperation` `PotentialEnergy`, defined as the potential energy per unit volume $E_{p} = \frac{g\rho z}{\rho_{0}}$, in a new script called `src/PotentialEnergyEquationTerms.jl`. `PotentialEnergy` can use either in-situ density or a user defined potential density for the calculation. To ensure `seawater_density` from Oceananigans.jl exists I updated the compat entry. As well, I needed to add [SeawaterPolynomials.jl](https://github.com/CliMA/SeawaterPolynomials.jl) as a dependency for the project. This has updated the `Manifest.toml` --- hope that is ok!

I added some validating functions similar to those already in use in the package and there are tests that check both the error throwing and a basic calculation to check things are running smoothly.

Closes #163 

Though this PR does not completely close #163 the background potential energy is something I will leave to another PR and refer back to the comments in #163 or open up a new issues where this can be kept track of.

### DIFF
--- a/src/DensityEquationTerms.jl
+++ b/src/DensityEquationTerms.jl
@@ -1,0 +1,120 @@
+module DensityEquationTerms
+
+using DocStringExtensions
+
+export GravitationalPotentialEnergy
+
+using Oceananigans.AbstractOperations: KernelFunctionOperation
+using Oceananigans: Models.seawater_density
+using Oceananigans: Models.model_geopotential_height
+using Oceananigans.Grids: Center, Face
+using Oceanostics: validate_location, validate_buoyancy
+
+
+"""
+    $(SIGNATURES)
+
+Return a `KernelFunctionOperation` to compute the`GravitationalPotentialEnergy`, `Eₚ = gρz`,
+at each grid `location` in `model`.
+**NOTE:** A `BoussinesqEquationOfState` must be used in the `model` to calculate
+`seawater_density`. See the [relevant documentation](https://clima.github.io/OceananigansDocumentation/dev/model_setup/buoyancy_and_equation_of_state/#Idealized-nonlinear-equations-of-state)
+for how to set `SeawaterBuoyancy` using a `BoussinesqEquationOfState`.
+
+Example
+=======
+
+By passing only the `model` to the function, the in-situ density is used in the calculation:
+```jldoctest
+julia> using Oceananigans, SeawaterPolynomials.TEOS10
+
+julia> using Oceanostics.DensityEquationTerms: GravitationalPotentialEnergy
+
+julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded))
+1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── Flat x
+├── Flat y
+└── Bounded  z ∈ [-1000.0, 0.0]   regularly spaced with Δz=10.0
+
+julia> tracers = (:T, :S)
+(:T, :S)
+
+julia> eos = TEOS10EquationOfState()
+BoussinesqEquationOfState{Float64}:
+    ├── seawater_polynomial: TEOS10SeawaterPolynomial{Float64}
+    └── reference_density: 1020.0
+
+julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos)
+SeawaterBuoyancy{Float64}:
+├── gravitational_acceleration: 9.80665
+└── equation_of_state: BoussinesqEquationOfState{Float64}
+
+julia> model = NonhydrostaticModel(; grid, buoyancy, tracers)
+NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
+├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── timestepper: QuasiAdamsBashforth2TimeStepper
+├── tracers: (T, S)
+├── closure: Nothing
+├── buoyancy: SeawaterBuoyancy with g=9.80665 and BoussinesqEquationOfState{Float64} with ĝ = NegativeZDirection()
+└── coriolis: Nothing
+
+julia> GravitationalPotentialEnergy(model)
+KernelFunctionOperation at (Center, Center, Center)
+├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── kernel_function: gravitational_potential_energy_ccc (generic function with 1 method)
+└── arguments: ("KernelFunctionOperation at (Center, Center, Center)", "KernelFunctionOperation at (Center, Center, Center)", "Tuple{Float64}")
+```
+
+To use a reference density pass the argument `reference_geopotential_height` to
+`GravitationalPotentialEnergy`:
+```jldoctest
+julia> using Oceananigans, SeawaterPolynomials.TEOS10
+
+julia> using Oceanostics.DensityEquationTerms: GravitationalPotentialEnergy
+
+julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded));
+
+julia> tracers = (:T, :S);
+
+julia> eos = TEOS10EquationOfState();
+
+julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos);
+
+julia> model = NonhydrostaticModel(; grid, buoyancy, tracers);
+
+julia> reference_geopotential_height = 0; # density variable will be σ₀
+
+julia> GravitationalPotentialEnergy(model, reference_geopotential_height)
+KernelFunctionOperation at (Center, Center, Center)
+├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── kernel_function: gravitational_potential_energy_ccc (generic function with 1 method)
+└── arguments: ("KernelFunctionOperation at (Center, Center, Center)", "KernelFunctionOperation at (Center, Center, Center)", "Tuple{Float64}")
+```
+"""
+@inline function GravitationalPotentialEnergy(model; location = (Center, Center, Center))
+
+    validate_location(location, "GravitationalPotentialEnergy")
+    validate_buoyancy(model.buoyancy)
+
+    grid = model.grid
+    ρ = seawater_density(model) # in-situ model density
+    Z = model_geopotential_height(model)
+    g = tuple(model.buoyancy.model.gravitational_acceleration)
+
+    return KernelFunctionOperation{Center, Center, Center}(gravitational_potential_energy_ccc, grid, ρ, Z, g)
+end
+@inline function GravitationalPotentialEnergy(model, reference_geopotential_height; location = (Center, Center, Center))
+
+    validate_location(location, "GravitationalPotentialEnergy")
+    validate_buoyancy(model.buoyancy)
+
+    grid = model.grid
+    σ = seawater_density(model, geopotential_height = reference_geopotential_height) # potential density reference to `reference_geopotential_height`
+    Z = model_geopotential_height(model)
+    g = tuple(model.buoyancy.model.gravitational_acceleration)
+
+    return KernelFunctionOperation{Center, Center, Center}(gravitational_potential_energy_ccc, grid, σ, Z, g)
+end
+
+@inline gravitational_potential_energy_ccc(i, j, k, grid, density, Z, g) = g[1] * density[i, j, k] * Z[i, j, k]
+
+end # module
--- a/src/DensityEquationTerms.jl
+++ b/src/DensityEquationTerms.jl
@@ -67,9 +67,9 @@ KernelFunctionOperation at (Center, Center, Center)
 To use a reference density pass the argument `reference_geopotential_height` to
 `GravitationalPotentialEnergy`:
 ```jldoctest
-julia> using Oceananigans, SeawaterPolynomials.TEOS10
+julia> using Oceananigans, SeawaterPolynomials.TEOS10;
 
-julia> using Oceanostics.DensityEquationTerms: GravitationalPotentialEnergy
+julia> using Oceanostics.DensityEquationTerms: GravitationalPotentialEnergy;
 
 julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded));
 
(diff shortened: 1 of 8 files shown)


src/DensityEquationTerms.jl
```diff
@@ -0,0 +1,120 @@
...
+    validate_location(location, "GravitationalPotentialEnergy")
+    validate_buoyancy(model.buoyancy)
+
+    grid = model.grid
+    σ = seawater_density(model, geopotential_height = reference_geopotential_height) # potential density reference to `reference_geopotential_height`
+    Z = model_geopotential_height(model)
+    g = tuple(model.buoyancy.model.gravitational_acceleration)
+
+    return KernelFunctionOperation{Center, Center, Center}(gravitational_potential_energy_ccc, grid, σ, Z, g)
+end
```
tomchor, line 116: Can we merge these two into one function? Would something like the suggestion below work?

```suggestion
@inline function GravitationalPotentialEnergy(model, reference_geopotential_height=nothing; location = (Center, Center, Center))

    validate_location(location, "GravitationalPotentialEnergy")
    validate_buoyancy(model.buoyancy)

    grid = model.grid
    σ = seawater_density(model, geopotential_height = reference_geopotential_height) # potential density reference to `reference_geopotential_height`
    Z = model_geopotential_height(model)
    g = tuple(model.buoyancy.model.gravitational_acceleration)

    return KernelFunctionOperation{Center, Center, Center}(gravitational_potential_energy_ccc, grid, σ, Z, g)
end
```

src/DensityEquationTerms.jl
```diff
@@ -0,0 +1,120 @@
...
+
+@inline gravitational_potential_energy_ccc(i, j, k, grid, density, Z, g) = g[1] * density[i, j, k] * Z[i, j, k]
```
tomchor, line 118: Two comments here:

1. Should this be `g[3]` instead of `g[1]`?
2. In Oceanostics I'm trying to name small functions like these as the actual operation they perform. So I'd change the name from `gravitational_potential_energy_ccc()` to `gρz_ccc()` or something. Do you agree that it'd be more readable?

src/Oceanostics.jl
```diff
@@ -38,6 +44,14 @@ validate_location(location, type, valid_location=(Center, Center, Center)) =
 validate_dissipative_closure(closure) = error("Cannot calculate dissipation rate for $closure")
 validate_dissipative_closure(::AbstractScalarDiffusivity{<:Any, ThreeDimensionalFormulation}) = nothing
 validate_dissipative_closure(closure_tuple::Tuple) = Tuple(validate_dissipative_closure(c) for c in closure_tuple)
+
+validate_buoyancy(buoyancy::Nothing) = throw(ArgumentError("Cannot calculate gravitational potential energy without a buoyancy model."))
+validate_buoyancy(buoyancy::Buoyancy{<:BuoyancyTracer, g}) where g =
```
tomchor, line 49: It would be trivial to implement potential energy with `BuoyancyTracer` though, right?

test/runtests.jl
```diff
@@ -1,14 +1,17 @@
 using Test
-using CUDA
+#using CUDA
```
tomchor, line 2: Did you forget to uncomment this?

test/runtests.jl
```diff
@@ -320,6 +323,55 @@ function test_tracer_diagnostics(model)
...
+
+    ρ = Field(seawater_density(model, geopotential_height = reference_geopotential_height))
+    compute!(ρ)
+    Z = Field(model_geopotential_height(model))
+    compute!(Z)
+    g = model.buoyancy.model.gravitational_acceleration
+
+    true_value =  @. g * ρ.data * Z.data
+
+    @test isequal(Eₚ_field.data, true_value)
```
tomchor, line 371: Can we merge these into one function?

tomchor:
This is looking great! Made a few comments. I don't have strong feelings either way, but I wonder if it's easier to name it `PotentialEnergy`, which is less verbose. How often are we encountering non-gravitational potential energy anyway...

src/DensityEquationTerms.jl
jbisits: Yes we can. One reason I did not is that I want to use the default behaviour of [`seawater_density`](https://clima.github.io/OceananigansDocumentation/dev/appendix/library/#Oceananigans.Models.seawater_density-Tuple{Union{HydrostaticFreeSurfaceModel,%20NonhydrostaticModel}}) for the in-situ density, and passing `nothing` to `seawater_density`,

```julia
seawater_density(model; geopotential_height  = reference_geopotential_height) # i.e.seawater_density(model; geopotential_height = nothing)
```

will not work. Instead we could use a conditional

```julia
ρ = isnothing(reference_geopotential_height) ? seawater_density(model) : seawater_density(model, geopotential_height = reference_geopotential_height) 
```

and then just need one function. Would that be ok?

test/runtests.jl
jbisits: Yep, depends on how [merging the functions is done](https://github.com/tomchor/Oceanostics.jl/pull/164#discussion_r1536057285) but will update this to one function after a decision there.

src/DensityEquationTerms.jl
jbisits: Regarding 2. yes that is a great idea and I will update!

For 1.,

```julia
g = tuple(model.buoyancy.model.gravitational_acceleration)
```

is a `Tuple` with a single entry which is the value of `gravitational_acceleration`, so to access the scalar value I think it should be `g[1]`. Not very clean way but it was so I could get to work on `GPU`. Is there a more appropriate way to pass a scalar to a `KernelFunctionOperation`? Maybe use `const` or `local const`? Or perhaps I should rename to use a `NamedTuple` called `parameters`,

```julia
parameters = (g = model.buoyancy.model.gravitational_acceleration)
``` 

which more closely follows the current design in Oceanostics.jl?

jbisits:
> This is looking great! Made a few comments. I don't have strong feelings either way, but I wonder if it's easier to name it `PotentialEnergy`, which is less verbose. How often are we encountering non-gravitational potential energy anyway...

I think rename to `PotentialEnergy` is good, will make the change!

src/DensityEquationTerms.jl
jbisits: I think this is best solution

```julia
parameters = (g = model.buoyancy.model.gravitational_acceleration)
```

so have made the change.

src/Oceanostics.jl
jbisits: I think you are right! Once the other comments here have been addressed could look at whether it is better to do in a new PR or make the change in this one?

test/runtests.jl
jbisits: Actually this can be done with `kwarg` and conditional so have done so.

src/DensityEquationTerms.jl
jbisits: I ended up using your suggestion of adding the keyword argument 

```julia
@inline function PotentialEnergy(model; geopotential_height = model_geopotential_height(model), location = (Center, Center, Center))

    validate_location(location, "PotentialEnergy")
    validate_buoyancy(model.buoyancy)

    grid = model.grid
    ρ = seawater_density(model; geopotential_height)
    Z = model_geopotential_height(model)
    parameters = (g = model.buoyancy.model.gravitational_acceleration,)

    return KernelFunctionOperation{Center, Center, Center}(gρz_ccc, grid, ρ, Z, parameters)
end
```

The default behviour is still to use in-situ density in the calculation but setting `geopotential_height = value` in  `PotentialEnergy` will instead use a potential density reference to `value`.

glwagner:
My main comment is to recommend defining "potential energy" as "potential energy per unit volume" rather than unit mass. I think it's more appropriate for a Boussinesq code.

This requires knowledge of the reference density when using `SeawaterBuoyancy`.

src/DensityEquationTerms.jl
glwagner, line 1: How about `PotentialEnergyEquationTerms.jl` for the name of this file?

This isn't really a density equation (ie mass conservation) thing. Potential energy is linked to the momentum equations.

src/Oceanostics.jl
tomchor: Agreed

tomchor:
> My main comment is to recommend defining "potential energy" as "potential energy per unit volume" rather than unit mass. I think it's more appropriate for a Boussinesq code. So the definition would be
> 
> Ep=gρzρr
> 
> Where ρr is the reference density. With `TEOS10`, we know the reference density. For `LinearEquationOfState`, the user has to additionally provide it.

I agree with this. Normalizing by density would also make the units consistent with the TKE equation terms defined in this package.

jbisits:
@tomchor with @glwagner's [suggestion above](https://github.com/tomchor/Oceanostics.jl/pull/164#issuecomment-2016538117) about defining potential energy per unit volume I was not sure what to rename 

```julia
 gρz_ccc(i, j, k, grid, ρ, Z, p) = (p.g / p.ρ₀) * ρ[i, j, k] * Z[i, j, k]
```

to. Following your earlier suggestion of naming small functions as the operation they perform it could be `gρ₀⁻¹ρz_ccc` but that is a lot of symbols. Happy for it to be this just wanted to check first in case you had any other thoughts!

src/PotentialEnergyEquationTerms.jl
```diff
@@ -0,0 +1,113 @@
...
+    grid = model.grid
+    ρ = seawater_density(model; geopotential_height)
+    Z = model_geopotential_height(model)
+    parameters = (g = model.buoyancy.model.gravitational_acceleration,
+                  ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
+
+    return KernelFunctionOperation{Center, Center, Center}(gρz_ccc, grid, ρ, Z, parameters)
+end
+
+@inline gρz_ccc(i, j, k, grid, ρ, Z, p) = (p.g / p.ρ₀) * ρ[i, j, k] * Z[i, j, k]
```
tomchor, line 111: One possibility is to implicitly call $g\rho/\rho_0=g\prime$, that way the units work out:

```suggestion
@inline g′z_ccc(i, j, k, grid, ρ, Z, p) = (p.g / p.ρ₀) * ρ[i, j, k] * Z[i, j, k]
```

tomchor: I'm also fine with naming it `gρ₀⁻¹ρz_ccc` like you said, or  `gρz_ρ₀_ccc`, and let the underscore implicitly mean division. As long as it can be understood as a representation of the operation, I'm happy :)

jbisits: I like `g′z_ccc` because it is the most compact.

tomchor (approved):
LGTM

@jbisits feel free to merge this after tests pass if you think this PR is done. You should be able to do it now that you're a collaborator.

src/PotentialEnergyEquationTerms.jl
```diff
@@ -99,11 +102,12 @@ KernelFunctionOperation at (Center, Center, Center)
     grid = model.grid
     ρ = seawater_density(model; geopotential_height)
     Z = model_geopotential_height(model)
-    parameters = (g = model.buoyancy.model.gravitational_acceleration,)
+    parameters = (g = model.buoyancy.model.gravitational_acceleration,
+                  ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
```
glwagner, line 106: This line will fail unless you're using an equation of state from `SeawaterPolynomials`

glwagner:
Note that the function should be `potential_energy` rather than `PotentialEnergy` (there is no type called `PotentialEnergy`)

src/PotentialEnergyEquationTerms.jl
jbisits: Yep that is correct but `validate_buoyancy` will throw an error if anything other than a model with `SeawaterBuoyancy{<:BoussinesqEquationOfState}` is passed. I aim to add methods for models with `SeawaterBuoyancy{<:LinearEquationOfState}` and `BuoyacnyTracer` soon.

jbisits:
> Note that the function should be `potential_energy` rather than `PotentialEnergy` (there is no type called `PotentialEnergy`)

Yes I remember you mentioning this when working on `seawater_density` in Oceananigans.jl. Though it seems in this package the functions that return a `KernelFunctionOperation` are capitalised camel case so I went with that. Can be changed at some point but should probably do that for the whole package in one go.

tomchor:
Yes, Oceanostics breaks away from Julia's [style guide](https://docs.julialang.org/en/v1/manual/style-guide/#Style-Guide) in the naming of its diagnostics. There's no good reason for that to be case tbh. In the beginning I just thought it made the code more readable like that and (since I assumed I was going to be the main person using it, like my previous similar packages) I just went with it. As the package grows we should probably consider bringing the whole thing in line with the style guide, but that should probably be done in one go for the whole package like @jbisits mentioned.

PR #167, opened 2024-03-26
Add `PotentialEnergy` methods for `BuoyancyTracer` and `SeawaterBuoyancy{<:LinearEquationOfState}`

jbisits:
Currently `PotentialEnergy` only works if the `BuoyancyModel` has a `BoussinesqEquationOfState`. This PR extends `PotentialEnergy` to have methods for both `BuoyancyTracer` and `SeawaterBuoyancy{<:LinearEquationOfState}`. It does so by dispatching on the `BuoyancyModel`.

To calculate the potential energy per unit volume for `BuoyancyTracer` and `SeawaterBuoyancy{<:LinearEquationOfState}` I have used

$$
E_{p} = bz
$$

where $b$ is buoyancy (I think this is what it should be but let me know if not!). In the case of `BuoyancyTracer` $b$ is exactly the buoyancy tracer and in the case of `SeawaterBuoyancy{<:LinearEquationOfState}` I compute the buoyancy via

https://github.com/tomchor/Oceanostics.jl/blob/68007f19e61f13152e8a6efd9a568eb933d941b2/src/PotentialEnergyEquationTerms.jl#L22-L23

I am not sure if this is the optimal way to do things so please let me know if you know of better ways to do this!

Once the calculation of $E_{p}$ is confirmed I will add some calculation tests to check things are doing what they should similar to what is currently there for when `SeawaterBuoyancy{<:BoussinesqEquationOfState}`

https://github.com/tomchor/Oceanostics.jl/blob/68007f19e61f13152e8a6efd9a568eb933d941b2/test/runtests.jl#L345-L358

### DIFF
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -17,8 +17,8 @@ using SeawaterPolynomials: BoussinesqEquationOfState
 
 const NoBuoyancyModel = Union{Nothing, ShallowWaterModel}
 const BuoyancyTracerModel = Buoyancy{<:BuoyancyTracer, g} where g
-const BuoyancyLinearEOSModel = Buoyancy{<:SeawaterBuoyancy{FT, <:LinearEquationOfState, T, S}} where {FT, T, S}
-const BuoyancyBoussinesqEOSModel = Buoyancy{<:SeawaterBuoyancy{FT, <:BoussinesqEquationOfState, T, S}} where {FT, T, S}
+const BuoyancyLinearEOSModel = Buoyancy{<:SeawaterBuoyancy{FT, <:LinearEquationOfState, T, S} where {FT, T, S}, g} where {g}
+const BuoyancyBoussinesqEOSModel = Buoyancy{<:SeawaterBuoyancy{FT, <:BoussinesqEquationOfState, T, S} where {FT, T, S}, g} where {g}
 
 validate_gravity_unit_vector(gravity_unit_vector::NegativeZDirection) = nothing
 validate_gravity_unit_vector(gravity_unit_vector) =
@@ -110,7 +110,7 @@ julia> PotentialEnergy(model)
 KernelFunctionOperation at (Center, Center, Center)
 ├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
 ├── kernel_function: g′z_ccc (generic function with 1 method)
-└── arguments: ("KernelFunctionOperation at (Center, Center, Center)", "KernelFunctionOperation at (Center, Center, Center)", "(g=9.80665, ρ₀=1020.0)")
+└── arguments: ("KernelFunctionOperation at (Center, Center, Center)", "(g=9.80665, ρ₀=1020.0)")
 ```
 
 To use a reference density set a constant value for the keyword argument `geopotential_height`
@@ -143,7 +143,7 @@ KernelFunctionOperation at (Center, Center, Center)
                                  geopotential_height = model_geopotential_height(model))
 
     validate_location(location, "PotentialEnergy")
-    validate_gravity_unit_vector(model.buoyancy.gravity_unit_vector)
+    isnothing(model.buoyancy) ? nothing : validate_gravity_unit_vector(model.buoyancy.gravity_unit_vector)
 
     return PotentialEnergy(model, model.buoyancy, geopotential_height)
 end
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -8,8 +8,28 @@ using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans: Models.seawater_density
 using Oceananigans: Models.model_geopotential_height
 using Oceananigans.Grids: Center, Face
+using Oceananigans.BuoyancyModels: Buoyancy, BuoyancyTracer, SeawaterBuoyancy, LinearEquationOfState
+using Oceananigans.Models: NonhydrostaticModel, HydrostaticFreeSurfaceModel
 using Oceanostics: validate_location, validate_buoyancy
+using SeawaterPolynomials: BoussinesqEquationOfState
 
+const ModelsWithBoussinesqEOS = Union{NonhydrostaticModel{TS, E, A, G, T,
+                                    <:Buoyancy{<:SeawaterBuoyancy{FT, <:BoussinesqEquationOfState, Temp, Sal}},
+                                    R, SD, U, C, Φ, F, V, S, K, BG, P, BGC, I, AF},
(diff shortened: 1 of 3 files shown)


src/PotentialEnergyEquationTerms.jl
```diff
@@ -5,11 +5,22 @@ using DocStringExtensions
...
+using Oceanostics: validate_location
+using SeawaterPolynomials: BoussinesqEquationOfState
 
+const NoBuoyancyModel = Union{Nothing, ShallowWaterModel}
+const BuoyancyTracerModel = Buoyancy{<:BuoyancyTracer, g} where g
+const BuoyancyLinearEOSModel = Buoyancy{<:SeawaterBuoyancy{FT, <:LinearEquationOfState, T, S}} where {FT, T, S}
+const BuoyancyBoussinesqEOSModel = Buoyancy{<:SeawaterBuoyancy{FT, <:BoussinesqEquationOfState, T, S}} where {FT, T, S}
+
+linear_eos_buoyancy(model) = linear_eos_buoyancy(model.grid, model.buoyancy.model, model.tracers)
+linear_eos_buoyancy(grid, buoyancy, tracers) = KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, buoyancy, tracers)
```
tomchor, line 23: Here you're defining a new `KernelFunctionOperation` that wraps around `buoyancy_perturbationᶜᶜᶜ`, which then is called again in defining the `PotentialEnergy` kernel.

I think that has compilation complexity costs (when you nest too many operations sometimes the GPU can't compile it). Can we just use `buoyancy_perturbationᶜᶜᶜ` and define only one `KernelFunctionOperation`?

tomchor:
I think the math here is correct.

One issue (which I think was also relevant for https://github.com/tomchor/Oceanostics.jl/pull/164 but I didn't catch it then), is that this calculation assumes that `gravity_unity_vector==NegativeZDirection()`. That's fine, I don't think we need to expand this for an arbitrary `gravity_unity_vector` (although that'd probably be easy), but I think we should test for that and throw a helpful error otherwise.

src/PotentialEnergyEquationTerms.jl
jbisits: Yep sure! I have also changed 

```julia
Z = model_geopotential_height(model)
```

which was also a `KernelFunctionOperation` to `Zᶜᶜᶜ`.

The only other thing on this front is that `seawater_density` returns a `KernelFunctionOperation` which creates a nested operation. If you think better this can be changed would just require a little bit of a rework.

src/PotentialEnergyEquationTerms.jl
```diff
@@ -5,11 +5,24 @@ using DocStringExtensions
 export PotentialEnergy
 
 using Oceananigans.AbstractOperations: KernelFunctionOperation
-using Oceananigans: Models.seawater_density
-using Oceananigans: Models.model_geopotential_height
+using Oceananigans.Models: seawater_density
+using Oceananigans.Models: model_geopotential_height
```
glwagner, line 9: I didn't realize that Oceananigans had a `model_geopotential_height` 😆 

Why isn't this just "geopotential_height"? What's different between the "height" and the "model height" ?

src/PotentialEnergyEquationTerms.jl
```diff
@@ -18,16 +31,48 @@ Return a `KernelFunctionOperation` to compute the `PotentialEnergy` per unit vol
...
+julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,))
+NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
+├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── timestepper: QuasiAdamsBashforth2TimeStepper
+├── tracers: b
+├── closure: Nothing
+├── buoyancy: BuoyancyTracer with ĝ = NegativeZDirection()
+└── coriolis: Nothing
+
+julia> PotentialEnergy(model)
```
glwagner, line 67: This should be `potential_energy` not `PotentialEnergy`...

src/PotentialEnergyEquationTerms.jl
```diff
@@ -87,27 +132,56 @@ julia> model = NonhydrostaticModel(; grid, buoyancy, tracers);
...
+
+    grid = model.grid
+    b = model.tracers.b
+
+    return KernelFunctionOperation{Center, Center, Center}(bz_ccc, grid, b)
+end
+
+@inline bz_ccc(i, j, k, grid, b) = b[i, j, k] * Zᶜᶜᶜ(i, j, k, grid)
+
+@inline function PotentialEnergy(model, buoyancy_model::BuoyancyLinearEOSModel, geopotential_height)
```
glwagner, line 164:
```suggestion
@inline function potential_energy(model, buoyancy_model::BuoyancyLinearEOSModel, geopotential_height)
```

src/PotentialEnergyEquationTerms.jl
jbisits: I think it was because it gets the height from the model it is passed. It was added as a fallback for [`seawater_density`](https://github.com/CliMA/Oceananigans.jl/blob/fb9455bc10721a880d9911a39735e37f1c4961e0/src/Models/seawater_density.jl#L122-L125) here

https://github.com/CliMA/Oceananigans.jl/blob/fb9455bc10721a880d9911a39735e37f1c4961e0/src/Models/seawater_density.jl#L21-L24

I don't think there is any difference between "height" and "model height". The name could be changed to "geopotential_height".

src/PotentialEnergyEquationTerms.jl
jbisits: Yes I agree. As mentioned [here](https://github.com/tomchor/Oceanostics.jl/pull/164#issuecomment-2019231307) I think it would be best to change the whole package but I guess could start here.

src/PotentialEnergyEquationTerms.jl
jbisits: What do you think @tomchor? I could make the change here then update the other functions in a new PR. Or could open a new PR to change functions to all be snake case in line with julia style. Now is as good a time as ever!?

src/PotentialEnergyEquationTerms.jl
glwagner: We can use `Oceananigans.geopotential_height` to avoid name conflicts

glwagner: Ah hmm we also have `model_temperature`. Perhaps we can leave it for now if this is mostly just a default...

src/PotentialEnergyEquationTerms.jl
tomchor: Let's keep the current style in this PR. We can discuss changes in an issue and have a separate PR to do it all at once if necessary.

tomchor (approved):
LGTM 

thanks @jbisits !

PR #169, opened 2024-04-16
Fix mean velocities not being used in KE dissipation calculation

zhihua-zheng:
Added a function to remove user provided mean velocities from `model.velocities` in `KineticEnergyDissipationRate`.

### DIFF
--- a/src/Oceanostics.jl
+++ b/src/Oceanostics.jl
@@ -74,8 +74,6 @@ end
 #+++ Utils for removing mean fields
 using Oceananigans: prognostic_fields, HydrostaticFreeSurfaceModel
 using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
-import Base: -
--(a::ZeroField, b::ZeroField) = a
 """
     $(SIGNATURES)
 
--- a/src/Oceanostics.jl
+++ b/src/Oceanostics.jl
@@ -72,8 +72,10 @@ end
 #---
 
 #+++ Utils for removing mean fields
-using Oceananigans: prognostic_fields
+using Oceananigans: prognostic_fields, HydrostaticFreeSurfaceModel
 using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
+import Base: -
+-(a::ZeroField, b::ZeroField) = a
 """
     $(SIGNATURES)
 
@@ -82,6 +84,10 @@ Remove mean fields from the model resolved fields.
 function perturbation_fields(model; kwargs...)
 
     resolved_fields = prognostic_fields(model)
+    if model isa HydrostaticFreeSurfaceModel
+        resolved_fields = (; resolved_fields..., w=ZeroField())
+    end
+
     mean_fields = values(kwargs)
     # Removes mean fields only if mean field is provided
     pert_fields = NamedTuple{keys(resolved_fields)}(haskey(mean_fields, key) ?
--- a/src/Oceanostics.jl
+++ b/src/Oceanostics.jl
@@ -71,23 +71,24 @@ function add_background_fields(model)
 end
 #---
 
-#+++ Utils for removing mean velocities
+#+++ Utils for removing mean fields
+using Oceananigans: prognostic_fields
 using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
 """
     $(SIGNATURES)
 
-Remove mean velocities from the model resolved velocities.
+Remove mean fields from the model resolved fields.
 """
-function remove_mean_velocities(model, mean_velocities)
+function fields_without_means(model; kwargs...)
 
-    velocities = model.velocities
-    # Removes mean velocities only if mean velocity isn't ZeroField
-    perturbation_velocities = NamedTuple{keys(velocities)}((mean_velocities[key] isa ZeroField) ?
-                                                           val :
-                                                           val - mean_velocities[key]
-                                                           for (key,val) in zip(keys(velocities), velocities))
-    return merge(perturbation_velocities,
-                 model.tracers,
+    resolved_fields = prognostic_fields(model)
+    mean_fields = values(kwargs)
+    # Removes mean fields only if mean field is provided
+    perturbation_fields = NamedTuple{keys(resolved_fields)}(haskey(mean_fields, key) ?
+                                                           val - mean_fields[key] :
+                                                           val
+                                                           for (key,val) in zip(keys(resolved_fields), resolved_fields))
+    return merge(perturbation_fields,
                  model.auxiliary_fields,
                  biogeochemical_auxiliary_fields(model.biogeochemistry))
 end
--- a/src/Oceanostics.jl
+++ b/src/Oceanostics.jl
@@ -71,6 +71,28 @@ function add_background_fields(model)
 end
 #---
 
+#+++ Utils for removing mean velocities
+using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
+"""
+    $(SIGNATURES)
+
+Remove mean velocities from the model resolved velocities.
+"""
+function remove_mean_velocities(model, mean_velocities)
+
+    velocities = model.velocities
+    # Removes mean velocities only if mean velocity isn't ZeroField
+    perturbation_velocities = NamedTuple{keys(velocities)}((mean_velocities[key] isa ZeroField) ?
+                                                           val :
+                                                           val - mean_velocities[key]
+                                                           for (key,val) in zip(keys(velocities), velocities))
+    return merge(perturbation_velocities,
+                 model.tracers,
+                 model.auxiliary_fields,
+                 biogeochemical_auxiliary_fields(model.biogeochemistry))
+end
+#---
+
 #+++ A few utils for closure tuples:
 using Oceananigans.TurbulenceClosures: νᶜᶜᶜ
 
--- a/src/Oceanostics.jl
+++ b/src/Oceanostics.jl
@@ -79,16 +79,16 @@ using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
 
 Remove mean fields from the model resolved fields.
 """
-function fields_without_means(model; kwargs...)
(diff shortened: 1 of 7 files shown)


src/Oceanostics.jl
```diff
@@ -71,6 +71,28 @@ function add_background_fields(model)
...
 #---
 
+#+++ Utils for removing mean velocities
+using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
+"""
+    $(SIGNATURES)
+
+Remove mean velocities from the model resolved velocities.
+"""
+function remove_mean_velocities(model, mean_velocities)
```
tomchor, line 81: I think the name of this function is a bit misleading since it doesn't just _remove mean velocities_, it returns a `models`'s fields minus the `mean_velocities`.

I think we should either:

1. Extend Oceananigan's `fields` function with `from Oceananigans import: fields`, and then replace `remove_mean_velocities` with `fields`, therefore adding one method to it.
2. Call it something else that's descriptive, such as `fields_without_mean()`

If we do the latter we might change the call to `fields_without_means(model, args...)`, so that we can remove the mean from any field (`u`, `v`, `w`, `b`, `T`, etc).

tomchor (changes requested):
Could you please add a quick test making sure that `KineticEnergyDissipationRate` works well when `U`, `V`, or `W` are passed? Let me know if you're confused about how to do that and I'll help you.

tomchor:
Thank you so much @zhihua-zheng! I made a couple of small comments and it should be ready to merge after that.

zhihua-zheng:
> Owner



> 

I actually don't how to do that. Some instructions would be great!

tomchor:
> > Could you please add a quick test making sure that `KineticEnergyDissipationRate` works well when `U`, `V`, or `W` are passed? Let me know if you're confused about how to do that and I'll help you.
> 
> I actually don't how to do that. Some instructions would be great!

No problem. Can you let me know once you do the rest? Then I'll do the testing myself :)

src/Oceanostics.jl
zhihua-zheng: I will go with the second suggestion here. But I'm not sure how to implement variable number of arguments in this case. Currently the function takes one `NamedTuple` containing mean velocities. Any psuedo-code example?

tomchor: This is working test example that should give you all the tools you need:

```julia
julia> function test(model; kwargs...)
           for (variable, mean) in kwargs
              @info "Here is where you you remive the $mean from $variable in $model"
           end
       end
test (generic function with 1 method)

julia> test("bla", a=4)
[ Info: Here is where you you remive the 4 from a in bla
```

src/TKEBudgetTerms.jl
```diff
@@ -283,13 +283,14 @@ where ∂ⱼuᵢ is the velocity gradient tensor and Fᵢⱼ is the stress tenso
 function KineticEnergyDissipationRate(model; U=ZeroField(), V=ZeroField(), W=ZeroField(),
                                                location = (Center, Center, Center))
     validate_location(location, "KineticEnergyDissipationRate")
-    u, v, w = model.velocities
+    mean_velocities = (u=U, v=V, w=W)
+    model_fields = fields_without_means(model; mean_velocities...)
```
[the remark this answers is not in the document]
zhihua-zheng: Yeah, `perturbation_fields` is even better.

tomchor:
@zhihua-zheng I just realized I can't do the changes for you right now because you forked this repo and made the changes in your own forked repo, to which I don't have access. Can you give me access to your forked repo? That might make it easier for me to modify this PR and add tests.

I just gave you access to Oceanostics, so that in the future you can just create a new branch here every time you open a PR, which should be easier.

tomchor:
@zhihua-zheng included a couple of tests that should be enough. Basically it tests that `KineticEnergyDissipationRate` is calculating the correct value without providing an average `U`, and then that it calculates the correct value with `U` provided. You can see that adding a test simply means adding the necessary statements to the file `test/runtests.jl`.

I think at this point this test will fail because `viscosity(model.closure, model.diffusivity_fields)` can return either a scalar or a `Field`, and right now the equation I included there will only work for fields. So that probably needs to be changed. I'm going on time off for the next week, so I just wanted to post this here as it is since I'm you can probably modify that and figure out how to make tests pass from here :)

test/runtests.jl
```diff
@@ -212,11 +213,24 @@ function test_ke_dissipation_rate_terms(grid; model_type=NonhydrostaticModel, cl
...
+
+    true_ε = viscosity(model.closure, model.diffusivity_fields) * dudz^2
+    @test interior(ε_field, 3, 3, 3)[1] == true_ε
+
+    ε = KineticEnergyDissipationRate(model; U=Field(Average(model.velocities.u, dims=(1,2))))
     ε_field = compute!(Field(ε))
     @test ε isa AbstractOperation
     @test ε_field isa Field
 
+    @test interior(ε_field, 3, 3, 3)[1] == 0.0
```
tomchor, line 232: I think probably this change will be necessary for tests to pass.

```suggestion
    @test interior(ε_field, 3, 3, 3)[1] ≈ 0.0
```

zhihua-zheng:
Test passed.

tomchor (approved):
Nice. Thanks @zhihua-zheng 

The only thing I ask: before merging, can you run the tests on a GPU? Unfortunately this has to be done manually, but you can just get an interactive Casper GPU and run `]test` on the repo's root directory.

zhihua-zheng:
Tests also passed on GPU. Also made some changes to let they pass for stretched grid and `HydrostaticFreeSurfaceModel`.

src/Oceanostics.jl
```diff
@@ -72,8 +72,10 @@ end
 #---
 
 #+++ Utils for removing mean fields
-using Oceananigans: prognostic_fields
+using Oceananigans: prognostic_fields, HydrostaticFreeSurfaceModel
 using Oceananigans.Biogeochemistry: biogeochemical_auxiliary_fields
+import Base: -
+-(a::ZeroField, b::ZeroField) = a
```
glwagner, line 78: Add to Oceananigans?

zhihua-zheng: Yeah, if you anticipate that could be useful outside Oceanostics.

glwagner: It's also a generic principle not to extend methods for types in a package that doesn't "own" those types (it's called type piracy, because you are "pirating" the behavior of the types 😄 ). Basically people might reimplement or (for some reason) implement different behavior in Oceananigans sometime in the future. Then this package would conflict with Oceananigans and could become unuseable... so for the benefit of all, best to extend methods inside the package that owns the types involved.

Obviously a pretty minor situation here, but its a teaching moment

glwagner: https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy

zhihua-zheng: Lesson learned! I can add a small PR to Oceananigans, or avoid it by trimming the `mean_fields` beforehand:
```julia
mean_fields = NamedTuple(key => val for (key,val) in zip(keys(mean_fields), mean_fields) if !(val isa ZeroField))
```
Which is better?

glwagner: Simpler is

```julia
mean_fields = NamedTuple(key => val for (key, val) in pairs(mean_fields) if !(val isa ZeroField))
```

But

```julia
-(a::ZeroField, b::ZeroField) = a
```

seems simpler enough a little neater. Only question I have, do you need to worry about the types of `a` and `b`? Cause `ZeroField` has an eltype:

https://github.com/CliMA/Oceananigans.jl/blob/1ee317b9b6986436f18dd12d7fcdd7067d9234b7/src/Fields/constant_field.jl#L5

seems like to do define the arithmetic well you need to promote types or something

glwagner:
```julia
function Base.-(z1::ZeroField{T1, N1}, z2::ZeroField{T2, N2}) where {T1, T2, N1, N2}
    T = Base.promote_type(T1, T2)
    N = max(N1, N2)
    return ZeroField{T, N}()
end
```

and heck while you're at it why not do a bit more

```julia
import Base: -, +, *

for op in (-, +, *)
    @eval begin
        function $(op)(z1::ZeroField{T1, N1}, z2::ZeroField{T2, N2}) where {T1, T2, N1, N2}
            T = Base.promote_type(T1, T2)
            N = max(N1, N2)
            return ZeroField{T, N}()
        end
    end
end
```

I think that's the right syntax but not 100% sure

glwagner: `ZeroField` is implemented here: https://github.com/CliMA/Oceananigans.jl/blob/main/src/Fields/constant_field.jl#L5

zhihua-zheng: Thanks! I will give it a try.

tomchor: Just for completeness, this is being implemented in https://github.com/CliMA/Oceananigans.jl/pull/3567

tomchor:
@zhihua-zheng I just started registering a new Oceananigans version with the changes in https://github.com/CliMA/Oceananigans.jl/pull/3567 (0.90.14).

After it's online (I guess it takes a couple of hours) please update the `Manifest.toml` to the latest version. Also this only works with the latest Oceananigans now, right? If so please update the `Project.toml`'s compat entry for Oceananigans to `0.90.14` and up. After that (and after tests pass) we should be good to merge :)

tomchor:
If this has been tested in GPUs, feel free to merge!

tomchor:
@zhihua-zheng you had mentioned on Slack that you were getting weird results with `KineticEnergyDissipationRate` when compared to `IsotropicKineticEnergyDissipationRate`. Since you merged, I'm assuming you resolve that. Can you quickly summarize what was going on? (If it's relevant)

zhihua-zheng:
I updated Oceananigans, and it worked as expected.

tomchor:
Awesome. Thanks for the contribution :)

Issue #168, opened 2024-03-28
Should we change function names to match Julia's style guide?


tomchor:
The naming of functions in this code doesn't follow [Julia's style guide](https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/). Namely, our _user-facing_ diagnostic functions have capitalized camel-case names. I'm emphasizing user-facing because all the internal functions already follow the style guide. As an example:

https://github.com/tomchor/Oceanostics.jl/blob/f5fbff7333100f04e8f67bcca288d74e2562fbe8/src/TKEBudgetTerms.jl#L362-L370

While the guide recommends that functions be named with lowercase. So the above function would be

```julia
function kinetic_energy_forcing_term(model::NonhydrostaticModel)
    ...
end
```

This was brought up in https://github.com/tomchor/Oceanostics.jl/pull/164 and https://github.com/tomchor/Oceanostics.jl/pull/167 so I thought I'd open an issue to discuss it.

cc @glwagner @jbisits

tomchor:
There are two reasons why things are coded this way. First, when I started coding the package I saw we were probably gonna use long verbose names clearly distinguish each diagnostic. So reducing function name length by avoiding underscores seemed like a good idea.

Second, and more importantly, I thought that camel case made code more readable. To get some examples from the docs, from the user perspective things currently look like this:

```julia
Q = QVelocityGradientTensorInvariant(model)
∫χᴰ = Integral(TracerVarianceDissipationRate(model, :b))
Ri = RichardsonNumber(model, add_background=true)
PV = ErtelPotentialVorticity(model, add_background=true)
```
According to Julia's style guide, we'd be using lowercase with underscores, making things look like

```julia
Q = q_velocity_gradient_tensor_invariant(model)
∫χᴰ = Integral(tracer_variance_dissipation_rate(model, :b))
Ri = richardson_number(model, add_background=true)
PV = ertel_potential_vorticity(model, add_background=true)
```

Of course this is a subjective point, but I think it's much easier to read the code the way things are currently written, especially since the keyword arguments are written in lowercase as well.

I'm curious about other points of view, but I don't think there's much advantage in changing the function name style for the users. I think the advantage of switching lies mainly when getting contributions from other people. When modifying something in the code, new developers currently might get a bit confused about what is a function and what is a type since they expect the style guide to be followed. Thus following the guide would potentially make contributions easier.

[...]

PR #172, opened 2024-04-22
Add background potential energy computation

jbisits:
This PR adds functions to compute the `BackgroundPotentialEnergy`, BPE, (see #168 for discussion on function naming) during a `simulation`. It does so using the definition given in  [Winters et al. (1995)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/available-potential-energy-and-mixing-in-densitystratified-fluids/A45F1A40521FF0A0DC82BC705AD398DA),

$$
E_{b} = g\int_{V}\rho z^{*}\mathrm{d}V.
$$

To compute $E_{b}$ the buoyancy/density `Field` from a model is first reshaped into a 1D `Array` and then sorted. The  $z^{*}$ (for a density `Field`) is defined as,

$$
z^{*} = \frac{1}{A}\int_{\rho_{\mathrm{min}}}^{\rho_{\mathrm{max}}}\mathrm{d}V.
$$

To compute  $z^{*}$, each volume element from the `model.grid` is computed, reshaped into a 1D `Array`, sorted using the `sortperm` from the buoyancy/density sorting, cumulatively summed then divided by the horizontal area of the `model.grid`.

These computations are done in `OneDReferenceField` which returns two new fields, the sorted buoyancy/density `Field` and the $z^{*}$ `Field`. These returned `Field`s are on a grid with a vertical extent equal to the original `model.grid` but with total number of elements equal to the number of points in the domain (i.e. `model.grid.Nx * model.grid.Ny * model.grid.Nz`). These two new `Field`s are then used to compute the BPE per unit volume. Calling `Integral(Field(BackgroundPotentialEnergy))` will give the volume integrated $E_{b}$ defined above.

I have handled the sorting of buoyancy and density separately because in the case of buoyancy, the buoyancy in the sorted profile should decrease as $z*$ increases,

$$
z^{*} = \frac{1}{A}\int_{b_{\mathrm{max}}}^{b_{\mathrm{min}}}\mathrm{d}V,
$$

whereas the density should increase as $z*$ increases. If this is not clear (or wrong!) please let me know!

I have not tried this using a `GPU` but I think that these functions (`reshape` and `sort`) work with `CuArray`s but I might have some scalar indexing that will cause problems --- when I get a chance to I will try on a `GPU`. 

The tests I have handled are really only validations that `OneDReferenceField` is computing things and returning things in the correct way. If you have any other suggestions for tests let me know.

Examples of the sorting for buoyancy and density using random noise as input data are below (they should be able to be run provided this branch of Oceanostics.jl is being used).

```julia
using Oceananigans, GLMakie
using Oceananigans.Models: seawater_density
using SeawaterPolynomials: TEOS10EquationOfState
using Oceanostics.PotentialEnergyEquationTerms

grid = RectilinearGrid(size=(10, 10, 50), x=(-10, 10), y=(-10, 10), z=(-100, 0),
                       topology=(Bounded, Bounded, Bounded))

## Using buoyancy
model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,))
set!(model, b=randn(size(grid)))

buoyancy_reference_profile, z✶ = OneDReferenceField(model.tracers.b, rev = true)

fig1 = Figure(size = (1000, 500))
ax = Axis(fig1[1, 1], title = "x-z slice of buoyancy field",
            xlabel = "x",
            ylabel = "z")
x, z = xnodes(model.grid, Center()), znodes(model.grid, Center())
hm = heatmap!(ax, x, z, interior(model.tracers.b, :, 1, :), colormap = :balance)
Colorbar(fig1[1, 2], hm, label = "buoyancy")
ax2 = Axis(fig1[1, 3], title = "Buoyancy reference profile",
            xlabel = "buoyancy",
            ylabel = "z✶")
lines!(ax2, interior(buoyancy_reference_profile, 1, 1, :), interior(z✶, 1, 1, :))
```
buoyancy_reference_profile

```julia
## Using density
eos = TEOS10EquationOfState()
model = NonhydrostaticModel(; grid, buoyancy=SeawaterBuoyancy(equation_of_state=eos), tracers=(:S, :T))
set!(model, S=randn(size(grid)), T=randn(size(grid)))
ρ = Field(seawater_density(model))
compute!(ρ)

density_reference_profile, z✶ = OneDReferenceField(ρ)

fig2 = Figure(size = (1000, 500))
ax = Axis(fig2[1, 1], title = "x-z slice of density field",
            xlabel = "x",
            ylabel = "z")
x, z = xnodes(model.grid, Center()), znodes(model.grid, Center())
hm = heatmap!(ax, x, z, interior(ρ, :, 1, :), colormap = :dense)
Colorbar(fig2[1, 2], hm, label = "density")
ax = Axis(fig2[1, 3], title = "Density reference profile",
            xlabel = "Density",
            ylabel = "z✶")
lines!(ax, interior(density_reference_profile, 1, 1, :), interior(z✶, 1, 1, :))
```
density_reference_profile

cc @janzika

### DIFF
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -282,7 +282,7 @@ end
     parameters = (g = model.buoyancy.model.gravitational_acceleration,
                   ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
 
-    return KernelFunctionOperation{Center, Center, Center}(g′z✶_ccc, grid, sorted_density, z✶, parameters)
+    return KernelFunctionOperation{Center, Center, Center}(minus_bz✶_ccc, grid, sorted_density, z✶, parameters)
 end
 
 @inline minus_bz✶_ccc(i, j, k, grid, sorted_ρ::Field, z✶::Field, p::NamedTuple) = (p.g / p.ρ₀) * sorted_ρ[i, j, k] * z✶[i, j, k]
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -50,21 +50,9 @@ julia> using Oceananigans
 
 julia> using Oceanostics.PotentialEnergyEquationTerms: PotentialEnergy
 
-julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded))
-1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── Flat x
-├── Flat y
-└── Bounded  z ∈ [-1000.0, 0.0]   regularly spaced with Δz=10.0
+julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded));
 
-julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,))
-NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
-├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── timestepper: QuasiAdamsBashforth2TimeStepper
-├── advection scheme: Centered reconstruction order 2
-├── tracers: b
-├── closure: Nothing
-├── buoyancy: BuoyancyTracer with ĝ = NegativeZDirection()
-└── coriolis: Nothing
+julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,));
 
 julia> PotentialEnergy(model)
 KernelFunctionOperation at (Center, Center, Center)
@@ -80,34 +68,15 @@ julia> using Oceananigans, SeawaterPolynomials.TEOS10
 
 julia> using Oceanostics.PotentialEnergyEquationTerms: PotentialEnergy
 
-julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded))
-1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── Flat x
-├── Flat y
-└── Bounded  z ∈ [-1000.0, 0.0]   regularly spaced with Δz=10.0
+julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded));
 
-julia> tracers = (:T, :S)
-(:T, :S)
+julia> tracers = (:T, :S);
 
-julia> eos = TEOS10EquationOfState()
-BoussinesqEquationOfState{Float64}:
-    ├── seawater_polynomial: TEOS10SeawaterPolynomial{Float64}
-    └── reference_density: 1020.0
+julia> eos = TEOS10EquationOfState();
 
-julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos)
-SeawaterBuoyancy{Float64}:
-├── gravitational_acceleration: 9.80665
-└── equation_of_state: BoussinesqEquationOfState{Float64}
+julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos);
 
-julia> model = NonhydrostaticModel(; grid, buoyancy, tracers)
-NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
-├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── timestepper: QuasiAdamsBashforth2TimeStepper
-├── advection scheme: Centered reconstruction order 2
-├── tracers: (T, S)
-├── closure: Nothing
-├── buoyancy: SeawaterBuoyancy with g=9.80665 and BoussinesqEquationOfState{Float64} with ĝ = NegativeZDirection()
-└── coriolis: Nothing
+julia> model = NonhydrostaticModel(; grid, buoyancy, tracers);
 
 julia> PotentialEnergy(model)
 KernelFunctionOperation at (Center, Center, Center)
@@ -257,20 +226,9 @@ julia> using Oceananigans
 
 julia> using Oceanostics.PotentialEnergyEquationTerms: BackgroundPotentialEnergy
 
-julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded))
-1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── Flat x
-├── Flat y
-└── Bounded  z ∈ [-1000.0, 0.0]   regularly spaced with Δz=10.0
+julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded));
 
-julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,))
-NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
-├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── timestepper: QuasiAdamsBashforth2TimeStepper
-├── tracers: b
-├── closure: Nothing
-├── buoyancy: BuoyancyTracer with ĝ = NegativeZDirection()
-└── coriolis: Nothing
+julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,));
 
 julia> bpe = BackgroundPotentialEnergy(model)
 KernelFunctionOperation at (Center, Center, Center)
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -12,7 +12,7 @@ using Oceananigans.Grids: NegativeZDirection
 using Oceananigans.BuoyancyModels: Buoyancy, BuoyancyTracer, SeawaterBuoyancy, LinearEquationOfState
 using Oceananigans.BuoyancyModels: buoyancy_perturbationᶜᶜᶜ, Zᶜᶜᶜ
 using Oceananigans.Models: ShallowWaterModel
-using Oceananigans.Fields: Field
+using Oceananigans.Fields: Field, compute!
 using Oceanostics: validate_location
 using SeawaterPolynomials: BoussinesqEquationOfState
 
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -2,7 +2,7 @@ module PotentialEnergyEquationTerms
 
 using DocStringExtensions
 
-export PotentialEnergy, BackgroundPotentialEnergy
+export PotentialEnergy, BackgroundPotentialEnergy, OneDReferenceField
 
 using Oceananigans.AbstractOperations: KernelFunctionOperation, volume, Az, GridMetricOperation
 using Oceananigans.Models: seawater_density
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -7,7 +7,7 @@ export PotentialEnergy, BackgroundPotentialEnergy, OneDReferenceField
 using Oceananigans.AbstractOperations: KernelFunctionOperation, volume, Az, GridMetricOperation
 using Oceananigans.Models: seawater_density
 using Oceananigans.Models: model_geopotential_height
-using Oceananigans.ImmersedBoundaryGrid: ImmersedBoundaryGrid
+using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid
 using Oceananigans.Grids
 using Oceananigans.Grids: Center, NegativeZDirection, interior, CenterField
 using Oceananigans.BuoyancyModels: Buoyancy, BuoyancyTracer, SeawaterBuoyancy, LinearEquationOfState
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -195,6 +195,7 @@ end
 
 """
     $(SIGNATURES)
+
 Return a `kernelFunctionOperation` to compute the `BackgroundPotentialEnergy` per unit
 volume,
 ```math
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -202,6 +202,7 @@ the smallest values are at the _bottom_ of the grid.
 
     return KernelFunctionOperation{Center, Center, Center}(sorted_field, grid, sf)
 end
+
 """
     function sort_field_revesre(f)
 Same method as [`sort_field`](@ref) but the 1D array is sorted in _descending_ order.
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -239,7 +239,7 @@ Examples
 ```jldoctest
 julia> using Oceananigans
 
-julia> using Oceanostics.PotentialEnergyEquationTerms: PotentialEnergy
+julia> using Oceanostics.PotentialEnergyEquationTerms: BackgroundPotentialEnergy
 
 julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded))
 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
@@ -260,7 +260,7 @@ julia> bpe = BackgroundPotentialEnergy(model)
 KernelFunctionOperation at (Center, Center, Center)
 ├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
 ├── kernel_function: bz_ccc (generic function with 2 methods)
-└── arguments: ("1×1×100 Array{Float64, 3}",)
+└── arguments: ("KernelFunctionOperation at (Center, Center, Center)",)
 ```
 """
 @inline function BackgroundPotentialEnergy(model; location = (Center, Center, Center),
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -3,6 +3,7 @@ module PotentialEnergyEquationTerms
 using DocStringExtensions
 
 export PotentialEnergy
+export sort_field, sorted_field
 
 using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans.Models: seawater_density
@@ -12,6 +13,7 @@ using Oceananigans.Grids: NegativeZDirection
 using Oceananigans.BuoyancyModels: Buoyancy, BuoyancyTracer, SeawaterBuoyancy, LinearEquationOfState
 using Oceananigans.BuoyancyModels: buoyancy_perturbationᶜᶜᶜ, Zᶜᶜᶜ
 using Oceananigans.Models: ShallowWaterModel
+using Oceananigans.Fields: Field
 using Oceanostics: validate_location
 using SeawaterPolynomials: BoussinesqEquationOfState
 
@@ -184,4 +186,23 @@ end
 
 @inline g′z_ccc(i, j, k, grid, ρ, p) = (p.g / p.ρ₀) * ρ[i, j, k] * Zᶜᶜᶜ(i, j, k, grid)
 
+## Background Potential Energy drafts
+
+resorted_field(i, j, k, grid, sf) = sf[i, j, k]
+function sorted_field(f::Field)
+    grid = f.grid
+    sorted_f = reshape(sort(reshape(f, :)), size(f))
+    return KernelFunctionOperation{Center, Center, Center}(resorted_field, grid, sorted_f)
+end
+
+function background_potential_energy(model)
+
+    return nothing
+end
+
+## Background potential energy would look like
+# Eb = gρ✶z/ρ₀ where ρ✶ is the sorted density field
+# or Eb = gρz✶/ρ₀ where z✶ = z[sortperm(ρ)]
+# so could have g/ρ₀ * ρ✶[i, j, k] * zᶜᶜᶜ(i, j, k, grid)
+# how to minimise the computation here? ie can I just return one kernel function operation?
 end # module
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -217,7 +217,7 @@ end
 Return a `KernelFunctionOperation` to compute the `BackgroundPotentialEnergy`
 per unit volume,
 ```math
-E_{b} = \\frac{gρz✶}{ρ₀} = -bz.
+E_{b} = \\frac{gρz✶}{ρ₀} = -bz✶.
 ```
 The `BackgroundPotentialEnergy` is the potential energy computed by adiabatically resorting
 the buoyancy or density field into a reference state of minimal potential energy.
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -40,9 +40,9 @@ and `SeawaterBuoyancy`. See the relevant Oceananigans.jl documentation on
 [buoyancy models](https://clima.github.io/OceananigansDocumentation/dev/model_setup/buoyancy_and_equation_of_state/)
 for more information about available options.
 
-The optional keyword argument `geopotential_height` is only used
-if ones wishes to calculate `Eₚ` with a potential density referenced to `geopotential_height`,
-rather than in-situ density, when using a `BoussinesqEquationOfState`.
+The keyword argument `geopotential_height` allows users to choose a set a different potential
+density for use in the calulation of `Eₚ` when using a `BoussinesqEquationOfState`. See the
+example below for how to do this.
 
 Example
 =======
@@ -64,8 +64,9 @@ KernelFunctionOperation at (Center, Center, Center)
 └── arguments: ("1×1×100 Field{Center, Center, Center} on RectilinearGrid on CPU",)
 ```
 
-The default behaviour of `PotentialEnergy` uses the *in-situ density* in the calculation
-when the equation of state is a `BoussinesqEquationOfState`:
+The default behaviour of `PotentialEnergy` uses *potential density* with reference
+`geopotential_height = 0` (i.e. σ₀ is the density variable) in the calculation when the
+equation of state is a `BoussinesqEquationOfState`:
 ```jldoctest
 julia> using Oceananigans, SeawaterPolynomials.TEOS10
 
@@ -88,8 +89,8 @@ KernelFunctionOperation at (Center, Center, Center)
 └── arguments: ("KernelFunctionOperation at (Center, Center, Center)", "(g=9.80665, ρ₀=1020.0)")
 ```
 
-To use a reference density set a constant value for the keyword argument `geopotential_height`
-and pass this the function. For example,
+To use a different potential density set a constant value for the keyword argument
+`geopotential_height` and pass this the function. For example,
 ```jldoctest
 julia> using Oceananigans, SeawaterPolynomials.TEOS10;
 
@@ -105,9 +106,9 @@ julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos);
 
 julia> model = NonhydrostaticModel(; grid, buoyancy, tracers);
 
-julia> geopotential_height = 0; # density variable will be σ₀
+julia> geopotential_height = 1000; # density variable will be σ₁
 
-julia> PotentialEnergy(model)
+julia> PotentialEnergy(model; geopotential_height)
 KernelFunctionOperation at (Center, Center, Center)
 ├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
 ├── kernel_function: g′z_ccc (generic function with 1 method)
@@ -150,7 +151,7 @@ end
 @inline function PotentialEnergy(model, buoyancy_model::BuoyancyBoussinesqEOSModel, geopotential_height)
 
     grid = model.grid
-    ρ = seawater_density(model; geopotential_height)
+    ρ = seawater_density(model; geopotential_height) # default to σ₀, potential density referenced to sea surface height
     parameters = (g = model.buoyancy.model.gravitational_acceleration,
                   ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
 
@@ -212,7 +213,7 @@ end
 """
     $(SIGNATURES)
 
-Return a `kernelFunctionOperation` to compute an approximation to the `BackgroundPotentialEnergy`
+Return a `KernelFunctionOperation` to compute the `BackgroundPotentialEnergy`
 per unit volume,
 ```math
 E_{b} = \\frac{gρz✶}{ρ₀}.
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -210,14 +210,14 @@ and is computed by cumulatively summing the 1D `Array` of grid volumes `ΔV`.
 **Note:** the `OneDReferenceField` is only appropriate for grids that have uniform horizontal
 area.
 """
-function OneDReferenceField(f::Field)
+function OneDReferenceField(f::Field; rev = false)
 
     area = sum(AreaField(f.grid))
     volume_field = VolumeField(f.grid)
     v = reshape(interior(volume_field), :)
     field_data = reshape(interior(f), :)
 
-    p = sortperm(field_data)
+    p = sortperm(field_data; rev)
     sorted_field_data = field_data[p]
     z✶ = cumsum(v[p]) / area
 
@@ -240,13 +240,12 @@ end
 Return a `kernelFunctionOperation` to compute the `BackgroundPotentialEnergy` per unit
 volume,
 ```math
-E_{b} = \\frac{gρ✶z}{ρ₀}.
+E_{b} = \\frac{gρz✶}{ρ₀}.
 ```
 The `BackgroundPotentialEnergy` is the potential energy computed after adiabatically resorting
 the buoyancy or density field into a reference state of minimal potential energy.
 The reference state is computed by reshaping the gridded buoyancy or density field and
-`sort`ing into a monotonically increasing `Vector`. This `sort`ed vector is then reshaped into
-the `size(model.grid)`.
+`sort`ing into a monotonically increasing `Vector`.
 
 Examples
 ========
@@ -274,8 +273,8 @@ NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
 julia> bpe = BackgroundPotentialEnergy(model)
 KernelFunctionOperation at (Center, Center, Center)
 ├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── kernel_function: bz_ccc (generic function with 2 methods)
-└── arguments: ("KernelFunctionOperation at (Center, Center, Center)",)
+├── kernel_function: bz✶_ccc (generic function with 1 method)
+└── arguments: ("1×1×100 Field{Center, Center, Center} on RectilinearGrid on CPU", "1×1×100 Field{Center, Center, Center} on RectilinearGrid on CPU")
 ```
 """
 @inline function BackgroundPotentialEnergy(model; location = (Center, Center, Center),
@@ -290,10 +289,12 @@ end
 @inline function BackgroundPotentialEnergy(model, buoyancy_model::BuoyancyTracerModel, geopotential_height)
 
     grid = model.grid
-    b✶ = sort_field(model.tracers.b)
+    sorted_buoyancy, z✶ = OneDReferenceField(model.tracers.b)
 
-    return KernelFunctionOperation{Center, Center, Center}(bz_ccc, grid, b✶)
+    return KernelFunctionOperation{Center, Center, Center}(bz✶_ccc, grid, sorted_buoyancy, z✶)
 end
+
+@inline bz✶_ccc(i, j, k, grid, b, z✶) = b[i, j, k] * z✶[i, j, k]
 
 linear_eos_buoyancy(grid, buoyancy, tracers) = KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, buoyancy, tracers)
 
@@ -303,20 +304,24 @@ linear_eos_buoyancy(grid, buoyancy, tracers) = KernelFunctionOperation{Center, C
     buoyancy = model.buoyancy.model
     tracers = model.tracers
     b = linear_eos_buoyancy(grid, buoyancy, tracers)
-    b✶ = sort_field(b)
+    compute!(b)
+    sorted_buoyancy, z✶ = OneDReferenceField(b)
 
-    return KernelFunctionOperation{Center, Center, Center}(bz_ccc, grid, b✶)
+    return KernelFunctionOperation{Center, Center, Center}(bz✶_ccc, grid, sorted_buoyancy, z✶)
 end
 
 @inline function BackgroundPotentialEnergy(model, buoyancy_model::BuoyancyBoussinesqEOSModel, geopotential_height)
 
     grid = model.grid
     ρ = seawater_density(model; geopotential_height)
-    ρ✶ = sort_field_reverse(ρ)
+    compute!(ρ)
+    sorted_density, z✶ = OneDReferenceField(ρ)
     parameters = (g = model.buoyancy.model.gravitational_acceleration,
                   ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
 
-    return KernelFunctionOperation{Center, Center, Center}(g′z_ccc, grid, ρ✶, parameters)
+    return KernelFunctionOperation{Center, Center, Center}(g′z_ccc, grid, sorted_density, z✶, parameters)
 end
+
+@inline g′z✶_ccc(i, j, k, grid, ρ, z✶, p) = (p.g / p.ρ₀) * ρ[i, j, k] * z✶[i, j, k]
 
 end # module
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -176,8 +176,9 @@ VolumeField(grid, loc=(Center, Center, Center)) = MetricField(loc, grid, volume)
     function OneDReferenceField(f::Field)
 Return a `OneDReferenceField` of the data in `f` and the corresponding `z✶` coordinate.
 The gridded data is first reshaped into a 1D `Array` then sorted. Returned is a new `Field`
-of this sorted data as well as the `z✶` `Field` that are both on a grid that has the same vertical extent
-as the original field (`f.grid.Lz`) but with the resolution of the whole domain (`Nx * Ny * Nz`).
+of this sorted data as well as the `z✶` `Field` that are both on a one-dimensional grid that
+has the same vertical extent as the original field (`f.grid.Lz`) but with the same _total_
+number of points of the original domain (`Nx * Ny * Nz`).
 The `z✶` `Field` is defined as
 ```math
 \\frac{1}{A}\\int_{f\\mathrm{min}}^{f\\mathrm{max}} \\mathrm{d}V.
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -193,26 +193,11 @@ array throughout the grid such that the _largest_ values are at the top of the g
 the smallest values are at the _bottom_ of the grid.
 """
 @inline sort_field(f) = reshape(sort(reshape(f, :)), size(f))
-
-@inline function sort_field(f::Field)
-    grid = f.grid
-    sorted_field = sort_field(f)
-    return KernelFunctionOperation{Center, Center, Center}(resorted_field, grid, sorted_field)
-end
 """
     function sort_field_revesre(f)
 Same method as [`sort_field`](@ref) but the 1D array is sorted in _descending_ order.
 """
 @inline sort_field_reverse(f) = reshape(sort(reshape(f, :), rev = true), size(f))
-
-@inline function sort_field_reverse(f::Field)
-    grid = f.grid
-    sorted_field = sort_field_reverse(f)
-    return KernelFunctionOperation{Center, Center, Center}(resorted_field, grid, sorted_field)
-end
-
-# For testing and validating `sort`ed fields.
-@inline resorted_field(i, j, k, grid, sf) = sf[i, j, k]
 
 """
     $(SIGNATURES)
--- a/src/PotentialEnergyEquationTerms.jl
+++ b/src/PotentialEnergyEquationTerms.jl
@@ -2,8 +2,7 @@ module PotentialEnergyEquationTerms
 
 using DocStringExtensions
 
-export PotentialEnergy
-export sort_field, sorted_field
+export PotentialEnergy, BackgroundPotentialEnergy
 
 using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans.Models: seawater_density
@@ -13,7 +12,7 @@ using Oceananigans.Grids: NegativeZDirection
 using Oceananigans.BuoyancyModels: Buoyancy, BuoyancyTracer, SeawaterBuoyancy, LinearEquationOfState
 using Oceananigans.BuoyancyModels: buoyancy_perturbationᶜᶜᶜ, Zᶜᶜᶜ
 using Oceananigans.Models: ShallowWaterModel
-using Oceananigans.Fields: Field
+using Oceananigans.Fields: Field, compute!
 using Oceanostics: validate_location
 using SeawaterPolynomials: BoussinesqEquationOfState
 
@@ -186,23 +185,121 @@ end
 
 @inline g′z_ccc(i, j, k, grid, ρ, p) = (p.g / p.ρ₀) * ρ[i, j, k] * Zᶜᶜᶜ(i, j, k, grid)
 
-## Background Potential Energy drafts
-
-resorted_field(i, j, k, grid, sf) = sf[i, j, k]
-function sorted_field(f::Field)
+# For testing `sort`ed fields.
+@inline resorted_field(i, j, k, grid, sf) = sf[i, j, k]
+@inline function sort_field(f::Field)
     grid = f.grid
-    sorted_f = reshape(sort(reshape(f, :)), size(f))
-    return KernelFunctionOperation{Center, Center, Center}(resorted_field, grid, sorted_f)
+    sorted_field = reshape(sort(reshape(f, :)), size(f))
+    return KernelFunctionOperation{Center, Center, Center}(resorted_field, grid, sorted_field)
 end
 
-function background_potential_energy(model)
+"""
+    $(SIGNATURES)
+Return a `kernelFunctionOperation` to compute the `BackgroundPotentialEnergy` per unit
+volume,
+```math
+Eₚ = \\frac{gρ✶z}{ρ₀}.
+```
+The `BackgroundPotentialEnergy` is the potential energy computed after adiabatically resorting
+the buoyancy or density field into a reference state of minimal potential energy.
+The reference state is computed by reshaping the gridded buoyancy or density field and
+`sort`ing into a monotonically increasing `Vector`. This `sort`ed vector is then reshaped into
+the `size(model.grid)`.
 
-    return nothing
+Examples
+========
+
+```jldoctest
+julia> using Oceananigans
+
+julia> using Oceanostics.PotentialEnergyEquationTerms: PotentialEnergy
+
+julia> grid = RectilinearGrid(size=5, z=(-5, 0), topology=(Flat, Flat, Bounded))
+1×1×5 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── Flat x
+├── Flat y
+└── Bounded  z ∈ [-5.0, 0.0]      regularly spaced with Δz=1.0
+
+julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,))
+NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
+├── grid: 1×1×5 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── timestepper: QuasiAdamsBashforth2TimeStepper
+├── tracers: b
+├── closure: Nothing
+├── buoyancy: BuoyancyTracer with ĝ = NegativeZDirection()
+└── coriolis: Nothing
+
+julia> set!(model, b = randn(size(model.grid)));
+
+julia> interior(model.tracers.b, 1, 1, :)
+5-element view(::Array{Float64, 3}, 1, 1, 4:8) with eltype Float64:
+  1.6132888274716937
+ -0.7877032120116347
+ -0.9357749288520266
+  1.310757334669075
+  0.09514543268446216
+
+julia> bpe = BackgroundPotentialEnergy(model)
+KernelFunctionOperation at (Center, Center, Center)
+├── grid: 1×1×5 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
+├── kernel_function: bz_ccc (generic function with 2 methods)
+└── arguments: ("1×1×5 Array{Float64, 3}",)
+
+julia> bpe_field = Field(bpe);
+
+julia> compute!(bpe_field);
+
+julia> interior(bpe_field, 1, 1, :)
+5-element view(::Array{Float64, 3}, 1, 1, 4:8) with eltype Float64:
+  4.210987179834119
+  2.7569612420407217
+ -0.2378635817111554
+ -1.9661360020036125
+ -0.8066444137358468
+```
+"""
+@inline function BackgroundPotentialEnergy(model; location = (Center, Center, Center),
+                                                  geopotential_height = model_geopotential_height(model))
+
+    validate_location(location, "BackgroundPotentialEnergy")
+    isnothing(model.buoyancy) ? nothing : validate_gravity_unit_vector(model.buoyancy.gravity_unit_vector)
+
+    return BackgroundPotentialEnergy(model, model.buoyancy, geopotential_height)
 end
 
-## Background potential energy would look like
-# Eb = gρ✶z/ρ₀ where ρ✶ is the sorted density field
-# or Eb = gρz✶/ρ₀ where z✶ = z[sortperm(ρ)]
-# so could have g/ρ₀ * ρ✶[i, j, k] * zᶜᶜᶜ(i, j, k, grid)
-# how to minimise the computation here? ie can I just return one kernel function operation?
+@inline function BackgroundPotentialEnergy(model, buoyancy_model::BuoyancyTracerModel, geopotential_height)
+
+    grid = model.grid
+    b✶ = reshape(sort(reshape(model.tracers.b, :)), size(grid))
+
+    return KernelFunctionOperation{Center, Center, Center}(bz_ccc, grid, b✶)
+end
+
+linear_eos_buoyancy(grid, buoyancy, tracers) = KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, buoyancy, tracers)
+
+@inline function BackgroundPotentialEnergy(model, buoyancy_model::BuoyancyLinearEOSModel, geopotential_height)
+
+    grid = model.grid
+    buoyancy = model.buoyancy.model
+    tracers = model.tracers
+    b = linear_eos_buoyancy(grid, buoyancy, tracers)
+    b = Field(b)
+    compute!(b)
+    b✶ = reshape(sort(reshape(b, :)), size(grid))
+
+    return KernelFunctionOperation{Center, Center, Center}(bz_ccc, grid, b✶)
+end
+
+@inline function BackgroundPotentialEnergy(model, buoyancy_model::BuoyancyBoussinesqEOSModel, geopotential_height)
+
+    grid = model.grid
+    ρ = Field(seawater_density(model; geopotential_height))
+    compute!(ρ)
+    ρ✶ = reshape(sort(reshape(ρ, :)), size(grid))
+    parameters = (g = model.buoyancy.model.gravitational_acceleration,
+                  ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
+
+    return KernelFunctionOperation{Center, Center, Center}(g′z_ccc, grid, ρ✶, parameters)
+end
+
 end # module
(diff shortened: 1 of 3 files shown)


src/PotentialEnergyEquationTerms.jl
```diff
@@ -184,4 +185,123 @@ end
...
+
+@inline function sort_field(f::Field)
+
+    compute!(f)
+    grid = f.grid
+    sf = sorted_field(f)
+
+    return KernelFunctionOperation{Center, Center, Center}(sorted_field, grid, sf)
+end
+"""
```
tomchor, line 205:
```suggestion
end

"""
```

tomchor:
@jbisits thanks this is great!

I'll only have time to look at this more carefully next week, but a couple of comments that I'll say right now are:

1. You're right, I think it would be best to avoid `KernelFuncionOperation` nesting. I think this can be done quite easily, no?
2. The sorting procedure is only an approximation to compute the BPE, so I think this should be mentioned somewhere in the docstring. (Or maybe even reflected in the function name? `ApproximateBackgroundPotentialEnergy`?)
3. Even then, sorting is only valid when the grid is regular (i.e. all the grid cells have equal volume). So we should test for that and spit out a useful error otherwise. (Or if you're up for it, implement the appropriate integral-form calculation of `z*`, as it would be useful to have something that works on stretched grids.)

A quick note is that I'm surprised that a `KFO` of `reshape(sort(reshape(f, :)), size(f))` works out of the box! I'd have thought that, because it's a nonlocal calculation (sorting), we would need to some "massaging" here. Glad I was wrong :)

glwagner:
> A quick note is that I'm surprised that a KFO of reshape(sort(reshape(f, :)), size(f)) works out of the box! I'd have thought that, because it's a nonlocal calculation (sorting), we would need to some "massaging" here. Glad I was wrong :)

Probably `sort` just uses `getindex` and `length`.

Personally I think that `sort` of a `Field` should return `Field`, not `Array`. (For example, this will almost certainly throw a scalar indexing error on GPU since we can't use `getindex` naively there.)

I think we should be able to support this without too much difficulty, provided we only support dimension-by-dimension sorting. If we do actually need to support sorting of Fields that are reshaped, we have a bit more work to do with `ReshapedArray`. Nothing too hard but it might be nice to know if we really need to support that before implementing it...

glwagner:
A little bit more about that... consider

```julia
julia> grid = RectilinearGrid(size=(1, 2, 3), x=(0, 1), y=(0, 1), z=(0, 1))
1×2×3 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0) regularly spaced with Δx=1.0
├── Periodic y ∈ [0.0, 1.0) regularly spaced with Δy=0.5
└── Bounded  z ∈ [0.0, 1.0] regularly spaced with Δz=0.333333

julia> c = CenterField(grid)
1×2×3 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 1×2×3 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── boundary conditions: FieldBoundaryConditions
│   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 7×8×9 OffsetArray(::Array{Float64, 3}, -2:4, -2:5, -2:6) with eltype Float64 with indices -2:4×-2:5×-2:6
    └── max=0.0, min=0.0, mean=0.0

julia> rc = reshape(c, :)
6-element reshape(::Field{Center, Center, Center, Nothing, RectilinearGrid{Float64, Periodic, Periodic, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing}}, 6) with eltype Float64:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> @which sort(rc)
sort(v::AbstractVector; kws...)
     @ Base.Sort sort.jl:1489
```

This calls

https://github.com/JuliaLang/julia/blob/9107bd3a2db63865b8818597feda0728807da613/base/sort.jl#L1730

But `copymutable(rc)` is just

```julia
julia> Base.copymutable(rc)
6-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
```

because we haven't defined `similar` for `ReshapedArray{<:Any, <:Any, <:Field}`:

```julia
help?> Base.ReshapedArray
  No documentation found.

  Summary
  ≡≡≡≡≡≡≡

  struct Base.ReshapedArray{T, N, P<:AbstractArray, MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}}
```

Supporting `sort!` seems rather easy, possibly as easy as

```julia
sort!(f::Field; kw...) = sort!(interior(f); kw...)
```

Note this also works on GPU. For `sort` we can take the approach that `CUDA` takes:

```julia
sort(f::Field; kw...) = sort!(copy(f); kw...)
```

(we need to check that `copy` works correctly, and I'm not sure if we want to use `deepcopy` or something more complicated).

Then, if we don't need to support `ReshapedArray`, we are done with these two lines.

jbisits:
Thanks for the comments and suggestions @tomchor and @glwagner!

@glwagner that does look like a cleaner (and better) implementation for `sort`ing a `Field`. I tried quite a few ways (including [your suggestion](https://github.com/tomchor/Oceanostics.jl/issues/163#issuecomment-2013092993)) but I was often making things a little more difficult than they needed to be --- so I ended up using the simplest implementation I came up with. I will use your suggestions and come up with a better (and seemingly even simpler) method to `sort` a `Field`.

> I'm a little confused about the sorting. Don't you want to sort only in the vertical? In the example you posted, the sorted field not only has a vertical buoyancy gradient but also a left-right buoyancy gradient which seems arbitrary...

Yes the horizontal buoyancy gradient that appears in the figures is arbitrary but part of that arises from `sort`ing a field which is just random noise. In the case of a stratified fluid I think we would see approximately uniform density over each horizontal layer. For reference here is the excerpt from section 6 in [Winters et al. (1995)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/available-potential-energy-and-mixing-in-densitystratified-fluids/A45F1A40521FF0A0DC82BC705AD398DA) which I have used when approximating the background potential energy

Screenshot 2024-04-24 at 9 43 49 AM

jbisits:
> @jbisits thanks this is great!
> 
> I'll only have time to look at this more carefully next week, but a couple of comments that I'll say right now are:
> 
> 1. You're right, I think it would be best to avoid `KernelFuncionOperation` nesting. I think this can be done quite easily, no?
> 2. The sorting procedure is only an approximation to compute the BPE, so I think this should be mentioned somewhere in the docstring. (Or maybe even reflected in the function name? `ApproximateBackgroundPotentialEnergy`?)
> 3. Even then, sorting is only valid when the grid is regular (i.e. all the grid cells have equal volume). So we should test for that and spit out a useful error otherwise. (Or if you're up for it, implement the appropriate integral-form calculation of `z*`, as it would be useful to have something that works on stretched grids.)
> 
> A quick note is that I'm surprised that a `KFO` of `reshape(sort(reshape(f, :)), size(f))` works out of the box! I'd have thought that, because it's a nonlocal calculation (sorting), we would need to some "massaging" here. Glad I was wrong :)

No problem! In response to the questions above:

1. Yep I think so especially given @glwagner's comments above
2. Definitely! Certainly in the docstring and probably no harm in putting it in the function name as well.
3. Again definitely! In this PR I will add a validation check with an error message for non-uniform grids. Then, if/when I get to it I can try and code up a calculation of `z*`.

glwagner:
@jbisits did you try

```julia
sort!(f::Field; kw...) = sort!(interior(f); kw...)
sort(f::Field; kw...) = sort!(copy(f); kw...)
```

?

> In the case of a stratified fluid I think we would see approximately uniform density over each horizontal layer.

But wouldn't any horizontal structure be destroyed if we resort in the horizontal? Consider doing this for the global ocean. It's cold at the poles and warm on the equator. Resorting in the horizontal would put cold water at the north pole and warm water at the south pole. That doesn't make sense. We must be assuming horizontal homogeneity. Is this necessary?

jbisits:
> @jbisits did you try
> 
> ```julia
> sort!(f::Field; kw...) = sort!(interior(f); kw...)
> sort(f::Field; kw...) = sort!(copy(f); kw...)
> ```
> 
> ?

Not yet I will get to it soon.

> > In the case of a stratified fluid I think we would see approximately uniform density over each horizontal layer.
> 
> But wouldn't any horizontal structure be destroyed if we resort in the horizontal? Consider doing this for the global ocean. It's cold at the poles and warm on the equator. Resorting in the horizontal would put cold water at the north pole and warm water at the south pole. That doesn't make sense. We must be assuming horizontal homogeneity. Is this necessary?

I am not sure if it matters that horizontal structure is not maintained after resorting (sorry if I am not understanding something here) because using the sorted array we want to attain an approximation to the minimal potential energy state which can be attained through adiabatic rearrangement of fluid parcels (or volume elements in Boussinesq case).

For your example above, if we first reshape a 3D density field for the global ocean into a 1D array and then sort it we would see the lighter, warmer water around the equator at the start of the sorted array and the deep, dense bottom water at the end end of the sorted array (with other water masses in between).  Redistributing this sorted array throughout the original grid would then have the lightest waters (e.g. from the around the equator) at the top of the grid $(z=0)$ and the densest water at the bottom of the grid (with the other water masses distributed in between). The result should be a stably stratified density field which can the be used to compute an approximation to the minimal potential energy state after adiabatic rearrangement which is the background potential energy. 

As you point out this could potentially result in warm water from the equator at the poles which does not make sense but I am not sure if that matters as we just want the lightest water on top to give a minimal potential energy state.

In the case of some experiments I have been running (more along direct numerical simulation lines) I would like to be able first reshape a 3D density field into a 1D array, sort this, then redistribute throughout the original grid. We could add a keyword argument that only sorts over the vertical dimension of the domain though? So we would have two cases one that sorts the reshaped array and redistributes over the grid

```julia
function sort!(f::Field; vertical_only = false, kw...)
    sorted_1D_field = reshape(interior(f), :)
    sort!(sorted_1D_field)
    f.data = reshape(sorted_1D_field, size(interior(f)) # I know this wont work because data has the halos is different size is immutable but this is the idea
    return f
end
```

and another that just sorts over the vertical dimension (assuming vertical dimension is 3 in example below) of the domain

```julia
sort!(f::Field; vertical_only = true, kw...) = sort!(interior(f), dims = 3; kw...)
```

Again sorry if I have misunderstood something but this is how I have been thinking about the BPE and its implementation for experiments I have been running.

glwagner:
> I am not sure if it matters that horizontal structure is not maintained after resorting (sorry if I am not understanding something here) because using the sorted array we want to attain an approximation to the minimal potential energy state which can be attained through adiabatic rearrangement of fluid parcels (or volume elements in Boussinesq case).

> For your example above, if we first reshape a 3D density field for the global ocean into a 1D array and then sort it we would see the lighter, warmer water around the equator at the start of the sorted array and the deep, dense bottom water at the end end of the sorted array (with other water masses in between). Redistributing this sorted array throughout the original grid would then have the lightest waters (e.g. from the around the equator) at the top of the grid 
 and the densest water at the bottom of the grid (with the other water masses distributed in between). The result should be a stably stratified density field which can the be used to compute an approximation to the minimal potential energy state after adiabatic rearrangement which is the background potential energy.

It's easy to construct a counter example. If we simply flip the y coordinate, we end up with a different background energy state using this method. But there can only be one physical background energy state --- right?

More severely I think the background energy state cannot be a function of horizontal position. It can only be a function of `z`.

Possibly the solution is as simple as averaging the sorted state in the horizontal directions.

It's also important to understand that this only works because of the indexing convention that we use. In other words there is an implicit assumption in the code that the z-index is the last index. Somehow, it might be nicer to write the code so that this assumption is either more clear or so that the code would still work if the z-index were 1st or 2nd.



There's more at play here. The background energy state can only be a function of z. It cannot be a function of horizontal position.

glwagner:
> I think that's the best we can do numerically and that's what winters et al suggest in their papers.

In the Winters papers, the background state is only a function of z and not of horizontal directions. Shouldn't our background state also be 1D then?

tomchor:
> > I think that's the best we can do numerically and that's what winters et al suggest in their papers.
> 
> In the Winters papers, the background state is only a function of z and not of horizontal directions. Shouldn't our background state also be 1D then?

In the physical world the background state is a function of $z$ because water parcels can be infinitesimally small). In our case each "parcel" is a grid cell that we're not willing to further subdivide, so it's natural that a resorting won't get rid of all horizontal gradients.

Your point of forcing the numerical, discrete background state to also be a function of $z$ only is interesting, but I'm not sure modifying the sorting procedure is the way to go. Here are some thoughts in no particular order:

1. This sorting procedure has become standard in the literature and experience seems to point to the errors produced by this simple sorting generally being small. My own experience corroborates that. The example concocted here by @jbisits is in no way physical so it overestimates this error in relation to realistic buoyancy field distribution.
2. I'd be reluctant to add an averaging procedure whose effects on the calculations we haven't investigated before. So I think if we do decide that we want to average in the horizontal, we need to do some investigating first with "realistic" simulations (whatever we think that means).
3. The sorting computation was always just meant to be an approximation to the background state. I think if we want to be more precise with the calculation, the way to go is to use the actual definition of the background state, rather than modifying the sorting procedure:

image

I feel like the equation above will be less performant (I think it'll always scale with O($n^2$)), but we can understand that as price to pay for precision.

@glwagner thoughts?

glwagner:
I'm not suggesting to change the sorting. I'm saying the background state should be derived by 

1. Sorting as in the present example
2. Averaging the sorted field horizontally to obtain a background state independent of z

This algorithm produces identical background states even if the horizontal coordinate direciton is reversed for example.

glwagner:
@tomchor can you please share the literature on this sorting-based calculation? 

I'm also missing how the sorted field is used to compute the available potential energy. Do you compute the difference between the current state and the sorted state then?

glwagner:
Here's a way to think about it... the variable $z_\star$ is the reference / resting height for the water parcel that is _currently_ at $x, t$,

image

The sorting that we are doing is precisely that adiabatic rearrangement being referred to. Somehow in fact, the information we would need to compute $z_\star$ must be contained in the permutator (ie output of `sortperm`) if we were to compute that.

When we average the sorted field in the horizontal direction, we obtain the density profile $\rho_\star(z_\star)$. This is a finite volume expression of what's written in Winters et al. If we did not want to sort in the horizontal direction, we could perhaps further imagine that each parcel is stretched to fill the whole horizontal domain. This could generate a sorted field on a very fine vertical grid with dimension `Nx * Ny * Nz`.

I think once we get the sorted field we can compute $z_\star$ more easily than equation 11 in Winters et al. It seems like for each grid point / density value we have to find the vertical level in the reference profile associated with it. That gives us a 3D field $z_\star$? It's not quite N^2, I think that's `Nx * Ny * Nz^2` because we have to search in `z` for every grid point). Maybe with an extra `log(Nz)`. Probably doable...

glwagner:
Here's an idea that generalizes to a (horizontally) stretched grid...

1. Define the total cell density $R = \rho \Delta V$. Also define a field of cell volumes, $\Delta V$.
2. Compute the sorting permutation for $R$, and use it sort $R$ and $\Delta V$.
3. Average both $R$ and $\Delta V$ over the "horizontal dimensions". The purpose of this is to produce a background field with the same "resolution" as the primitive fields. However, they will be on different grids, so the next task is to compute the "grid" of the background field...
4. Compute $\rho_\star(z_k) = \bar R / \overline{\Delta V}$, and the height of each increment $\Delta z_k$ by $\Delta z_k = \overline{\Delta V}_k / A$ where $A$ is the area of the domain.

Would this work? Hmm, as I wrote this I realized this will not work for irregular domains. Anyways, it's a start perhaps.

tomchor:
> I'm not suggesting to change the sorting. I'm saying the background state should be derived by
> 
> 1. Sorting as in the present example
> 2. Averaging the sorted field horizontally to obtain a background state independent of z

I understand that. Sorry, I guess I'm expressing myself poorly. That's what I meant by "modifying the sorting procedure", which is admittedly a misnomer.
 
> This algorithm produces identical background states even if the horizontal coordinate direciton is reversed for example.
> 
> I'm hazy on the relationship between the sorted field and z⋆. Isn't the result of this computation to actually obtain the background field ρ(z⋆)? The additional step of computing z⋆ is actually what one needs to do in order to relate the computed background to physical space z... ?

The variable $z_*$ is the height at which a parcel would be if all parcels were adiabatically rearranged to the state of minimal PE. If the parcels can be made infinitesimal this is state is unique and there are no horizontal buoyancy gradients. Which is why the equation presented to calculate it (Eq (11) in their paper, which I reproduced [here](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2083672638)) has a unique answer. That equation should be the preferred way to calculate $z_\star$, but afaik people rarely use it because it's computationally expensive to calculate. The calculation to get the BPE goes like this:

1. From a field $b$ (or $\rho$) you use Equation (11) to get a field $z_\star$ (the height of each parcel if all parcels were adiabatically resorted to a state of total minimum PE).
2. You then multiply $b$ by $z_\star$ to get the PE of each parcel in this state of minimum total PE. Note that in this calculation never really resorted anything, as really we only need the height at which each parcel would sit, and Equation (11) gives us that.
3. Finally you integrate everything and then you have the total PE of the domain in a minimum PE state: the BPE. Note that, in the paper, most (all?) the calculations are done with domain-integrated quantities since the BPE is necessarily a nonlocal quantity; that's these calculation are always a bit of a hassle.

Now, if you wanna avoid calculating Equation (11), instead of manually calculating $z_\star$ we can, in the words of Winters et al., "let the fluid elements correspond to volume elements of a numerical grid" and actually sort buoyancy values. Then the vertical coordinate (originally $z$) _approximately_ becomes $z_\star$ by definition, since each parcel is _approximately_ where it would be if the values were resorted to a statement of minimum PE. I'm emphasizing "approximately" here because as we discussed, grid cells aren't infinitesimal, and their sorting only approximates the state of minimum PE, and their sorted state not only isn't unique, but in general has some small horizontal gradients.

- The advantage of this latter method is that is faster since we never explicitly calculate $z_*$; it implicitly appears after the sorting is complete. Basically we exchange a $O(n^2)$ operation (Equatoin (11)) for a $\approx O(n log(n))$ operation.
- The disadvantage is that it's only an approximation.

Then the calculation the get the BPE through this approach is:

1. You 3D sort a field of $b$.
2. After resorting the original vertical coordinate $z$ becomes $z_\star$. We ignore horizontal coordinates. Then we can just multiply the sorted $b$ by its coordinate (which now approximates $z_\star$) to obtain the PE of each parcel in the minimum PE state.
3. Integrate everything to get the (approximated) BPE.

My take is that the latter method (i.e. sorting) will always be an approximation to the former method (i.e. direct calculation of $z_\star$ with Eq. (11)). Thus we can use the sorting approach as it is until an example/simulation comes along where that approximation fails and we need something better. Then we can either implement the first method (direct calculation of $z_\star$) or try to improve on the approximated one.

@jbisits @glwagner thoughts?

glwagner:
> Then the vertical coordinate (originally ) approximately becomes  by definition, since each parcel is approximately where it would be if the values were resorted to a statement of minimum PE

I agree that the vertical coordinate of the sorted field is $z_\star$. But this does not give you the _function_ $f(x, t) = z_\star$ in terms of the unsorted coordinates $(x, t)$. Something is funky here...

tomchor:
> > Then the vertical coordinate (originally ) approximately becomes  by definition, since each parcel is approximately where it would be if the values were resorted to a statement of minimum PE
> 
> I agree that the vertical coordinate of the sorted field is z⋆. But this does not give you the _function_ f(x,t)=z⋆ in terms of the unsorted coordinates (x,t). Something is funky here...

I think you're right. It doesn't. But I think we don't need that function. Just $z_\star$.

glwagner:
Can you be more explicit? What exactly is the calculation that you do after you obtain the sorted field (which is the density field, not a coordinate)?

jbisits:
> 
> > This algorithm produces identical background states even if the horizontal coordinate direciton is reversed for example.
> > I'm hazy on the relationship between the sorted field and z⋆. Isn't the result of this computation to actually obtain the background field ρ(z⋆)? The additional step of computing z⋆ is actually what one needs to do in order to relate the computed background to physical space z... ?
> 
> The variable z∗ is the height at which a parcel would be if all parcels were adiabatically rearranged to the state of minimal PE. If the parcels can be made infinitesimal this is state is unique and there are no horizontal buoyancy gradients. Which is why the equation presented to calculate it (Eq (11) in their paper, which I reproduced [here](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2083672638)) has a unique answer. That equation should be the preferred way to calculate z⋆, but afaik people rarely use it because it's computationally expensive to calculate. The calculation to get the BPE goes like this:
> 
> 1. From a field b (or ρ) you use Equation (11) to get a field z⋆ (the height of each parcel if all parcels were adiabatically resorted to a state of total minimum PE).
> 2. You then multiply b by z⋆ to get the PE of each parcel in this state of minimum total PE. Note that in this calculation never really resorted anything, as really we only need the height at which each parcel would sit, and Equation (11) gives us that.
> 3. Finally you integrate everything and then you have the total PE of the domain in a minimum PE state: the BPE. Note that, in the paper, most (all?) the calculations are done with domain-integrated quantities since the BPE is necessarily a nonlocal quantity; that's these calculation are always a bit of a hassle.
> 
> Now, if you wanna avoid calculating Equation (11), instead of manually calculating z⋆ we can, in the words of Winters et al., "let the fluid elements correspond to volume elements of a numerical grid" and actually sort buoyancy values. Then the vertical coordinate (originally z) _approximately_ becomes z⋆ by definition, since each parcel is _approximately_ where it would be if the values were resorted to a statement of minimum PE. I'm emphasizing "approximately" here because as we discussed, grid cells aren't infinitesimal, and their sorting only approximates the state of minimum PE, and their sorted state not only isn't unique, but in general has some small horizontal gradients.
> 
> * The advantage of this latter method is that is faster since we never explicitly calculate z∗; it implicitly appears after the sorting is complete. Basically we exchange a O(n2) operation (Equatoin (11)) for a ≈O(nlog(n)) operation.
> * The disadvantage is that it's only an approximation.
> 
> Then the calculation the get the BPE through this approach is:
> 
> 1. You 3D sort a field of b.
> 2. After resorting the original vertical coordinate z becomes z⋆. We ignore horizontal coordinates. Then we can just multiply the sorted b by its coordinate (which now approximates z⋆) to obtain the PE of each parcel in the minimum PE state.
> 3. Integrate everything to get the (approximated) BPE.
> 
> My take is that the latter method (i.e. sorting) will always be an approximation to the former method (i.e. direct calculation of z⋆ with Eq. (11)). Thus we can use the sorting approach as it is until an example/simulation comes along where that approximation fails and we need something better. Then we can either implement the first method (direct calculation of z⋆) or try to improve on the approximated one.
> 
> @jbisits @glwagner thoughts?

Yes this is my understanding of how to obtain an approximation for the volume integrated BPE and is what I have tried to implement in this PR.

jbisits:
> > @tomchor can you please share the literature on this sorting-based calculation?
> 
> I don't think there's more literature specifically on the sorting-based calculation other than the original paper. Winter and D'Asaro touch on it again on their follow-up 1996 paper but I think that's it.
> 
> My impression that the sorting-based approach works well comes from reading papers that employed it without complications, plus talking to a few people that corroborate that impression. I'll try to remember some of the papers that I read which use that calculation and post here later (I don't remember any of them off the top of my head).

[Caulfield and Peltier (2000)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/anatomy-of-the-mixing-transition-in-homogeneous-and-stratified-free-shear-layers/CF5269CD6948EEE80F5219B5ACF4A4ED) and  [Carpenter et al. (2012)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/simulations-of-a-doublediffusive-interface-in-the-diffusive-convection-regime/63D2ECE2AA41439E01A01F9A0D76F2E2) compute the BPE of a 3D scalar field using a sorted 1D array. The excerpt from Carpenter et al. (2012) where they show how they do this is

Screenshot 2024-05-01 at 8 41 26 AM (2)

In this calculation they have computed an "equivalent" (not sure if this is useful terminology) $z$ coordinate, $z_{b}$, for the sorted 1D scalar array. I have also used this method but I thought that the most simple way to have the BPE implemented in Oceanostics was reshape the sorted scalar field back on to the original grid so that `Integral` could be called (as pointed out by @tomchor it is the integrated quantity we almost always, or even always, want). I think this is also similar to what @glwagner [wrote above](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2083852351)?

One way to calculate this "equivalent" $z$ coordinate is to:

1. compute the volume, $\Delta V$, of each grid cell over a 3D domain
2. reshape the gridded values of $\Delta V$ into a 1D array and cumulatively sum to get the amount of volume up to a given value of iso-scalar (i.e. a given value in the sorted 1D array of scalar values)
3. divide the cumulatively summed volume array by the horizontal surface area (if it is constant, otherwise divide each volume element by each horizontal area element , $\Delta v / \Delta A$) and you have an equivalent $z$ array.

This is similar to the methods developed in [Sohail et al (2021)](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2020GL091439) where they reshape a 3D scalar field into a 1D array and sort to look at the amount of a scalar quantity up to a fixed volume (or equivalent depth after dividing by surface area).

I have used this method to obtain an equivalent $z$ and then compute the integrated BPE. Comparing the results (for the very small number of cases I have looked at) using this method vs the method implemented in the PR there was negligible difference in the two values.

jbisits:
> Can you be more explicit? What exactly is the calculation that you do after you obtain the sorted field (which is the resting density field, not a coordinate mapping)?
> 
> For example, equation 19 in Winters et al. 1995 is
> 
> image
> For this we need the function f(x,t)=z⋆. How do we compute equation 19 without explicitly obtaining z⋆?

Equation (19) is for the available potential energy

$$
E_{a} = g \int_{V}\rho (z - z^{*}) \mathrm{d}V
$$ 

which in Winters et al. (1995) is defined as the difference between the potential energy

$$
E_{p} = g\int_{V} \rho z \mathrm{d}V
$$

and background potential energy

$$
E_{b} = g \int_{V}\rho z^{*} \mathrm{d}V.
$$

In our case, we can approximate $E_{b}$ by

$$
E_{b}' \approx g \int_{V}\rho^{*} z \mathrm{d}V,
$$

where $\rho^{*}$ is the sorted 1D array of the density field that has been reshaped back onto the original grid. The function to compute the potential energy is already in Oceanostics so we can then approximate the available potential energy equation by

$$
E_{a}' \approx E_{p} - E_{b}'
$$ 

and take a finite difference approximation of $E_{a}'$ to get an approximation of equation (19).

I believe this is the method suggested by Winters et al. (1995) in the section titled numerical implementation.

Is this similar to what you have used in the past @tomchor?

tomchor:
> Is this similar to what you have used in the past @tomchor?

Yes, that's pretty much it!

tomchor:
> Can you be more explicit? What exactly is the calculation that you do after you obtain the sorted field (which is the resting density field, not a coordinate mapping)?
> 
> For example, equation 19 in Winters et al. 1995 is
> image
> 
> For this we need the function f(x,t)=z⋆. How do we compute equation 19 without explicitly obtaining z⋆?

For the calculation of all the terms in that equation, yes, we need to explicitly calculate $z_\star$ (only due to the last term I think). But we're not trying to calculate that whole equation in this PR. For what we're trying to calculate (which was summarized neatly by @jbisits in https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2087722148 and amounts to approximating the BPE so that we can calculate the APE) an explicit calculation of $z_\star$ isn't needed (as long as you're okay with the approximation and its (hopefully small) errors).

glwagner:
> > Is this similar to what you have used in the past @tomchor?
> 
> Yes, that's pretty much it!

But how do you compute $z - z_\star$?

glwagner:
> [Caulfield and Peltier (2000)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/anatomy-of-the-mixing-transition-in-homogeneous-and-stratified-free-shear-layers/CF5269CD6948EEE80F5219B5ACF4A4ED) and [Carpenter et al. (2012)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/simulations-of-a-doublediffusive-interface-in-the-diffusive-convection-regime/63D2ECE2AA41439E01A01F9A0D76F2E2) compute the BPE of a 3D scalar field using a sorted 1D array.

This clarifies things a lot and makes much more sense to me than reshaping the sorted field back to the dimensions of the 3D field (which doesn't make sense to me, or rather I think its wrong based on the arguments I gave above).

jbisits:
> > > Is this similar to what you have used in the past @tomchor?
> > 
> > 
> > Yes, that's pretty much it!
> 
> But how do you compute z−z⋆? The math is not the problem, it's the code that is missing. I don't see how you are doing it after obtaining the sorted density field.

You do not have to compute $z - z*$ when using this method. We compute $E_{p}$ and approximate $E_{b}$ and take a difference which gives an approximation to the APE without calculating $z - z*$ (this is method proposed by Winters et al. (1995) which we are following). As @tomchor pointed out the point of sorting the density field is

>After resorting the original vertical coordinate $z$ becomes $z^*$

So using the sorted density field (I have called this $\rho*$ [above](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2087722148)) and the original vertical coordinate $z$ we approximate $\rho z*$ with $\rho^* z$ and then use $\rho^* z$ in the approximation for $E_{b}$. This means we can approximate the two quantities $E_{p}$ and $E_{b}$ and take the difference to approximate $E_{a}$ all without calculating $z - z*$.

jbisits:
> > [Caulfield and Peltier (2000)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/anatomy-of-the-mixing-transition-in-homogeneous-and-stratified-free-shear-layers/CF5269CD6948EEE80F5219B5ACF4A4ED) and [Carpenter et al. (2012)](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/simulations-of-a-doublediffusive-interface-in-the-diffusive-convection-regime/63D2ECE2AA41439E01A01F9A0D76F2E2) compute the BPE of a 3D scalar field using a sorted 1D array.
> 
> This clarifies things a lot and makes much more sense to me than reshaping the sorted field back to the dimensions of the 3D field (which doesn't make sense to me, or rather I think its wrong based on the arguments I gave above).
> 
It is just a different method to approximate the background potential energy. It can be implemented like this it just means we need to define and compute $z{b}$ (to use Carpenter et al.'s terminology) and there are probably a few ways to do this.

> > Comparing the results (for the very small number of cases I have looked at) using this method vs the method implemented in the PR there was negligible difference in the two values.
> 
> Isn't it easy to come up with an example that breaks it?
> 
> It's dangerous to write code we know is wrong. Sometimes it ends up being more effort than the time saved by making the approximation, because we have to laboriously explain when the code is wrong and give people advice about when / when not to use it, not to mention losing sleep at night...

I am not quite sure this indicates that I have written code here that I know is wrong. I was saying that so far I have used  several methods (e.g. the method presented in this PR and something similar to Carpenter et al. (2012) with two different versions of the $z_{b}$) to obtain an estimate for the volume integrated background potential energy (BPE). The estimates were all in agreement to around 9 decimal places. I know this is not perfect and perhaps, or even likely, there is an example that breaks it. The point I was trying to get across is there is more than one method to approximate the BPE.

The method I chose to implement in this PR is based off the original suggestion by Winters et al. (1995) and is the method that, after some experiments and discussions with my supervisors, we chose to use. For stratified fluids on uniform grids it should produce a reasonable approximation to the BPE.

glwagner:
> > > > Is this similar to what you have used in the past @tomchor?
> > > 
> > > 
> > > Yes, that's pretty much it!
> > 
> > 
> > But how do you compute z−z⋆? The math is not the problem, it's the code that is missing. I don't see how you are doing it after obtaining the sorted density field.
> 
> You do not have to compute z−z∗ when using this method. We compute Ep and approximate Eb and take a difference which gives an approximation to the APE without calculating z−z∗ (this is method proposed by Winters et al. (1995) which we are following). As @tomchor pointed out the point of sorting the density field is
> 
> > After resorting the original vertical coordinate z becomes z∗
> 
> So using the sorted density field (I have called this ρ∗ [above](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2087722148)) and the original vertical coordinate z we approximate ρz∗ with ρ∗z and then use ρ∗z in the approximation for Eb. This means we can approximate the two quantities Ep and Eb and take the difference to approximate Ea all without calculating z−z∗.

Ok, thank you for the extended implementation (I got confused in your original explanation and didn't see what you wrote below). And pointing to Winters et al. 1995 is helpful --- I agree they make exactly the same argument you have been making.

The notation is a little confusing, but I don't think there is any approximation between what you define as $E_b$ and $E'_b$. It's just a change of coordinates from physical coordinate system to the "resorted coordinate system". Since $dV$ are the same between the two, you can integrate the density in the resorted coordinate system just fine.

Winters et al. 1995 state that the approximate part is the _reshaping_, since the reference profile should be 1D but they (like you) are using a 3D state that isn't the true minimum potential energy state (because fluid elements with different densities are side by side rather than vertically sorted).

I'm sort of scratching my head though --- isn't it quite trivial to compute the true reference density, at least on a uniform grid? After going to all of the huge amount of effort to write a CFD code and do all of these computations, we really stop short of the endzone at this point? It doesn't make sense to me. We simply have to sum the 1D sorted field in a straightforward manner... ?

If the grid is stretched things are indeed more complicated, because we have to take into account the different volumes of the elements that contribute to the reference profile. I think what I proposed aabove is a solution though. Not that hard, at least in a domain that doesn't have immersed boundaries... or am I missing something?

jbisits:
> I'm sort of scratching my head though --- isn't it quite trivial to compute the true reference density, at least on a uniform grid? After going to all of the huge amount of effort to write a CFD code and do all of these computations, we really stop short of the endzone at this point? It doesn't make sense to me. We simply have to sum the 1D sorted field in a straightforward manner... ?

Yep that it is totally true!

> If the grid is stretched things are indeed more complicated, because we have to take into account the different volumes of the elements that contribute to the reference profile. I think what I proposed above is a solution though. Not that hard, at least in a domain that doesn't have immersed boundaries... or am I missing something?

Nope, I do not think it should be too hard. I will implement a form of this over the next days.

glwagner:
> > I'm sort of scratching my head though --- isn't it quite trivial to compute the true reference density, at least on a uniform grid? After going to all of the huge amount of effort to write a CFD code and do all of these computations, we really stop short of the endzone at this point? It doesn't make sense to me. We simply have to sum the 1D sorted field in a straightforward manner... ?
> 
> Yep that it is totally true!
> 
> > If the grid is stretched things are indeed more complicated, because we have to take into account the different volumes of the elements that contribute to the reference profile. I think what I proposed above is a solution though. Not that hard, at least in a domain that doesn't have immersed boundaries... or am I missing something?
> 
> Nope, I do not think it should be too hard. I will implement a form of this over the next days.

Nice! Doing this right seems like a lot of payoff for the effort (I guess Julia helps).

Extending to irregular domains can remain an open challenge, I have an inkling that somebody, someday will want to tackle that.

tomchor:
> > Isn't it easy to come up with an example that breaks it?
> > It's dangerous to write code we know is wrong. Sometimes it ends up being more effort than the time saved by making the approximation, because we have to laboriously explain when the code is wrong and give people advice about when / when not to use it, not to mention losing sleep at night...
> 
> I am not quite sure this indicates that I have written code here that I know is wrong. I was saying that so far I have used several methods (e.g. the method presented in this PR and something similar to Carpenter et al. (2012) with two different versions of the zb) to obtain an estimate for the volume integrated background potential energy (BPE). The estimates were all in agreement to around 9 decimal places. I know this is not perfect and perhaps, or even likely, there is an example that breaks it. The point I was trying to get across is there is more than one method to approximate the BPE.

I agree with @jbisits here. The code isn't wrong. Both sorting procedures (1D and 3D) are approximations to D'Asaro's Equation (11).

> Winters et al. 1995 state that the approximate part is the _reshaping_, since the reference profile should be 1D but they (like you) are using a 3D state that isn't the true minimum potential energy state (because fluid elements with different densities are side by side rather than vertically sorted).

Can you point to (or post) the statement where they say that here? The way I understand it, resorting everything with finite, grid-size, indivisible volumes is a source of error from the start.

> > > I'm sort of scratching my head though --- isn't it quite trivial to compute the true reference density, at least on a uniform grid? After going to all of the huge amount of effort to write a CFD code and do all of these computations, we really stop short of the endzone at this point? It doesn't make sense to me. We simply have to sum the 1D sorted field in a straightforward manner... ?
> > 
> > 
> > Yep that it is totally true!
> > > If the grid is stretched things are indeed more complicated, because we have to take into account the different volumes of the elements that contribute to the reference profile. I think what I proposed above is a solution though. Not that hard, at least in a domain that doesn't have immersed boundaries... or am I missing something?
> > 
> > 
> > Nope, I do not think it should be too hard. I will implement a form of this over the next days.
> 
> Nice! Doing this right seems like a lot of payoff for the effort (I guess Julia helps).
> 
> Extending to irregular domains can remain an open challenge, I have an inkling that somebody, someday will want to tackle that.

After all this back and forth I'm a bit confused. What are you calling the "right" approach here? IMO the "right" way to do things (i.e. the un-approximated way) is implementing their [Equation (11)](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2083672638) and directly calculating $z_\star$. That approach can be used for both regular and stretched grids, as well as immersed boundary grids.

jbisits:
> After all this back and forth I'm a bit confused. What are you calling the "right" approach here? IMO the "right" way to do things (i.e. the one with the fewest approximations) is implementing their [Equation (11)](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2083672638) and directly calculating z⋆. That approach can be used for both regular and stretched grids, as well as immersed boundary grids.

Yes I think that is correct --- the most accurate method is implementing $z*$. Perhaps we could go with your initial suggestion of putting approximate into the function name (e.g. `ApproximateBackgroudPotentialEnergy`) then look to implement $z*$ in another PR?

glwagner:
> I agree with @jbisits here. The code isn't wrong. Both sorting procedures (1D and 3D) are approximations to D'Asaro's Equation (11).

It's just semantics. An approximation is "wrong", by definition. That's what I meant.

> Can you point to (or post) the statement where they say that here?

The statement is here:

image

Doing an exact computation merely means generating a sorted density field that is parallel to gravity (and doing that correctly). It's not very hard...

> IMO the "right" way to do things (i.e. the one with the fewest approximations) is implementing their [Equation (11)](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2083672638)

The only approximation is the reshaping. If we generate the reference state correctly, then we can evaluate the integrated background potential energy. This does not require explicitly computing $z_\star$.

However, if one wishes to compute $z_\star$ for other reasons (if we need more than just the _integrated_ background potential energy), then more work is required. Equation (11) is one way, but I think there are more efficient ways as I have outlined, that would introduce an additional fact only in the vertical direction (rather than computing a 3D integral at every point, as equation 11 does).

tomchor:
Okay I think I finally understand what @glwagner is proposing. Correct me if I'm wrong but what you want to do is to deform (flatten, really) each grid cell when sorting, so that they'd go from having a volume $\Delta v_i = \Delta x \Delta y \Delta z_i$ to having volume $\Delta v_f = Lx Ly \Delta z_f$. That way they can be stacked on top of each other (via a 1D sorting) and the final result will be a sorted domain with the same dimensions we started with. I guess that's what Carpenter et al. (2012) were describing [here](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2087702090), but it definitely wasn't clear to me when reading that passage. Pretty ingenious! And I agree that's better.

A note is that while this can be generalized to stretched grids if you're careful with with spacings as @glwagner mentioned, it doesn't generalize to `ImmersedGrid`s or (I think) cases where gravity isn't aligned with the grid's `z` direction. For that (plus computing other terms in the APE equation) we'd need a direct calculation of $z_\star$, which we probably should implement in another PR.

I think we're finally all on the same page now! :rocket:

glwagner:
> A note is that while this can be generalized to stretched grids if you're careful with with spacings as @glwagner mentioned, it doesn't generalize to `ImmersedGrid`s or (I think) cases where gravity isn't aligned with the grid's `z` direction. For that (plus computing other terms in the APE equation) we'd need a direct calculation of z⋆, which we probably should implement in another PR.
> 
> I think we're finally all on the same page now! 🚀

Agree! Getting this right for complex domains requires some extra work. But it's ok to leave a few things for the next generation.

It might good to throw error when gravity direction is not z or when using immersed boundary grid, also to help people who may be interested in contributing that extension.

tomchor:
> It might good to throw error when gravity direction is not z or when using immersed boundary grid, also to help people who may be interested in contributing that extension.

For sure!

src/PotentialEnergyEquationTerms.jl
```diff
@@ -50,21 +50,9 @@ julia> using Oceananigans
...
-├── grid: 1×1×100 RectilinearGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo
-├── timestepper: QuasiAdamsBashforth2TimeStepper
-├── advection scheme: Centered reconstruction order 2
-├── tracers: b
-├── closure: Nothing
-├── buoyancy: BuoyancyTracer with ĝ = NegativeZDirection()
-└── coriolis: Nothing
+julia> grid = RectilinearGrid(size=100, z=(-1000, 0), topology=(Flat, Flat, Bounded));
+
+julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=(:b,));
```
glwagner, line 59:
```suggestion
julia> model = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=:b);
```

src/PotentialEnergyEquationTerms.jl
```diff
@@ -186,4 +161,137 @@ end
...
 
+## Grid metrics from https://github.com/tomchor/Oceanostics.jl/issues/163#issuecomment-2012623824
+
+function MetricField(loc, grid, metric)
+
+    metric_operation = GridMetricOperation(loc, metric, grid)
+    metric_field = Field(metric_operation)
+
+    return compute!(metric_field)
+end
```
tomchor, line 172:
```suggestion
function MetricField(loc, grid, metric)
    metric_operation = GridMetricOperation(loc, metric, grid)
    metric_field = Field(metric_operation)
    return compute!(metric_field)
end
```

src/PotentialEnergyEquationTerms.jl
```diff
@@ -186,4 +161,137 @@ end
...
+
+VolumeField(grid, loc=(Center, Center, Center)) = MetricField(loc, grid, volume)
+  AreaField(grid, loc=(Center, Center, Nothing)) = MetricField(loc, grid, Az)
+
+"""
+    function OneDReferenceField(f::Field)
+Return a `OneDReferenceField` of the data in `f` and the corresponding `z✶` coordinate.
+The gridded data is first reshaped into a 1D `Array` then sorted. Returned is a new `Field`
+of this sorted data as well as the `z✶` `Field` that are both on a grid that has the same vertical extent
+as the original field (`f.grid.Lz`) but with the resolution of the whole domain (`Nx * Ny * Nz`).
```
tomchor, line 182: @jbisits is my suggestion correct? I think it's clearer that way.

```suggestion
Return a `OneDReferenceField` of the data in `f` and the corresponding `z✶` coordinate.
The gridded data is first reshaped into a 1D `Array` then sorted. Returned is a new `Field`
of this sorted data as well as the `z✶` `Field` that are both on a one-dimensional grid that
has the same vertical extent as the original field (`f.grid.Lz`) but with the same _total_
number of points of the original domain (`Nx * Ny * Nz`).
```

src/PotentialEnergyEquationTerms.jl
```diff
@@ -186,4 +161,137 @@ end
...
+
+    grid = model.grid
+    ρ = seawater_density(model; geopotential_height)
+    compute!(ρ)
+    sorted_density, z✶ = OneDReferenceField(ρ)
+    parameters = (g = model.buoyancy.model.gravitational_acceleration,
+                  ρ₀ = model.buoyancy.model.equation_of_state.reference_density)
+
+    return KernelFunctionOperation{Center, Center, Center}(g′z_ccc, grid, sorted_density, z✶, parameters)
+end
```
tomchor, line 293: I know we usually prefer multiple dispatch but there's a lot of overlap between these four functions. Should we put everything into one or two functions and just use plain old if/else statements?

glwagner: "multiple dispatch" is not the reason why you can't achieve code re-use here. `if` statements are fine too though, I just think you want to think clearly about the changes that are needed.

jbisits: Yeah sure! I did it like this because I thought it would make it easy to see how the computations are done for the different `Buoyancy` models. I will try and simplify.

jbisits:
I am still in the process of getting this finished but do keep comments as will likely speed things up. Once it is ready I will write up everything in the PR [intro box](https://github.com/tomchor/Oceanostics.jl/pull/172#issue-2255463122) (does this have a better name?).

src/PotentialEnergyEquationTerms.jl
```diff
@@ -186,4 +159,126 @@ end
...
+    set!(z✶_field, reshape(z✶, size(new_grid)))
+
+    return sorted_field, z✶_field
+
+end
+
+"""
+    $(SIGNATURES)
+
+Return a `kernelFunctionOperation` to compute an approximation to the `BackgroundPotentialEnergy`
```
tomchor, line 215: I think with the algorithm with discussed, this isn't an approximation anymore, correct?

hdrake:
Just a note that @ikeshwani has been independently computing BPE, for now just by brute-force offline for the non-immersed case eventually handling the `ImmersedBoundary` case. @liuchih is probably interested in computing BPE with tilted gravity vector and/or on domains that are periodic in the direction of the gravity vector (e.g. where there is a stratified background buoyancy field).

We'll be keeping a close eye on these developments and hope to contribute to generalizations in these more complicated contexts.

src/PotentialEnergyEquationTerms.jl
jbisits: I thought the reshaping was an approximation but I think you might be right. Will have a double check now.

hdrake:
@jbisits, have you tested the new diagnostics on anything more complicated than a 1D profile? @ikeshwani implemented a nice [notebook](https://github.com/hdrake/HorizontalConvection/blob/6075279ca01b53764c569f5158a79b27be77b39f/src/shear_instability_simulation.ipynb) that qualitatively replicates the Winters et al. (1995) shear instability example, which you might find useful.

jbisits:
> @jbisits, have you tested the new diagnostics on anything more complicated than a 1D profile and a random 2D field? @ikeshwani implemented a nice [notebook](https://github.com/hdrake/HorizontalConvection/blob/6075279ca01b53764c569f5158a79b27be77b39f/src/shear_instability_simulation.ipynb) that qualitatively replicates the Winters et al. (1995) shear instability example, which you might find useful.

Not the diagnostics in this PR (I have used `PotentialEnergy` in a DNS) but it would be great to use notebook you have shared and this branch to look the results are. I will get to it soon - Thanks very much for sharing this!!!

src/PotentialEnergyEquationTerms.jl
tomchor: It definitely was with the simple resorting we started with. But if you start with a distribution that's discrete, like we are since this is a numerical model on a grid, then I think reshaping doesn't introduce any further approximations, so I think we're okay removing the "approximation" from the dosctring. But I 100% certain. Just... 95% certain haha.

tomchor:
> It could be a good addition to the documentation once this PR is finished if @ikeshwani is interested?

Agreed! To me it sounds like it'd be great to merge @ikeshwani's notebook with the [Kelvin-Helmholtz example](https://tomchor.github.io/Oceanostics.jl/stable/generated/kelvin_helmholtz/) that's already in the docs. That example performs some very simple diagnostics, and it would be great to use that to reproduce results from Winters et al. (1995) in addition to what's already there.

src/PotentialEnergyEquationTerms.jl
jbisits: Will do!

hdrake:
I've implemented an very efficient offline algorithm for computing $z^{star}$ with the sorting method [[in this notebook](https://github.com/hdrake/HorizontalConvection/blob/abf3c5e0da4e0c521f613c2ddd95ca6b681c6e2a/notebooks/background_potential_energy.ipynb)], that in principle handles both variable grid spacing and immersed boundaries. I don't yet understand the kernel techniques required to implement this GPUs well enough myself, but I think it should be obvious from the notebook how to extend @jbisits' current diagnostic to handle these more general configurations.

Two modifications to the algorithm are required to generalize the method to support immersed boundaries:
1) *Ocean area changes with depth:* The vertical `cumsum` to get `zstar` needs to be done via a for-loop because the area needs to be updated *as you go up in the water column* and the horizontal ocean area at the sorted depth changes. I do this by creating an interpolation function for $A(z)$ and then interpolating to every $z^{star}$, but that may be overkill and a nearest-neighbor look-up could suffice.
2) *Skipping land cells when iterating through grid cells:* By default, `sort` and the more useful `sortperm` functions throw `NaN` values to the end of the sorted array, so I used this to essentially skip over the land cells. Also, when you map the sorted buoyancy field (as a 1D vector) back into the 3D array, you need to be careful to skip the land cells!

tomchor:
> I've finally caught up on this whole thread and want to clarify a few things:
> ### Regarding approximations
> 
> > The sorting procedure is only an approximation to compute the BPE, so I think this should be mentioned somewhere in the docstring. (Or maybe even reflected in the function name? ApproximateBackgroundPotentialEnergy?)
> 
> @tomchor, could you clarify what the approximation being made is? My understanding is that the method is only approximate for equations of state that depend on pressure, and are exact for linear EOS or a buoyancy tracer. Some negligible error might be introduced if the discretization method does not align perfectly with the model's grid discretization, but that is a fundamentally different thing than saying we are making a physical approximation to the Boussinesq BPE. (Or maybe you are saying that Boussinesq BPE is already an approximate to non-Boussinesq BPE?) But the physics of our diagnostics should be consistent with our model physics.

At the time we were planning to do a simple 1D sorting, and the reshape everything back to 3D, exactly like described in the original paper. This would incur horizontal gradients, which is what I was talking about. Since then, after discussion, we decided to do a 1D sorting only the vertical and flatten the cell volumes, without any 3D reshaping (you can follow the previous comments for more details). With this, there's no further approximating other than the ones you mention.

> ### Regarding irregular grids
> 
> > Even then, sorting is only valid when the grid is regular (i.e. all the grid cells have equal volume). So we should test for that and spit out a useful error otherwise. (Or if you're up for it, implement the appropriate integral-form calculation of z*, as it would be useful to have something that works on stretched grids.)
> 
> @tomchor, the sorting method is easily generalized to account for unequal grid cell volumes; you just have to account for the actual volume of a parcel when you stack up the elevation increments to compute z⋆. Maybe this is what you had in mind, but you don't actually need to evaluate the integrals of the heaviside directly, which is an prohibitively expensive O(N2) operation.

Agreed. Again, this comment was before we decided to do a 1D sorting along with cell flattening. Note that we still probably want to implement the expensive $z_\star$ calculation at some point since it's still needed for tilted domains or domains with immersed boundaries. I wouldn't say it's _prohibitively_ expensive. But yes, definitely expensive since it scales as $N^2$. (Although at some point you mentioned that it can scale as $nlog(n)$? I don't see how but hopefully you're right!)

hdrake:
> Note that we still probably want to implement the expensive $z^{\star}$ calculation at some point since it's still needed for tilted domains or domains with immersed boundaries.

I don't think that's true. I've already implemented it for immersed boundaries and I think the method could be generalized to tilted domains but that one is more complicated because of the background buoyancy field and boundary conditions.

src/PotentialEnergyEquationTerms.jl
```diff
@@ -169,9 +143,9 @@ end
 
     grid = model.grid
     C = model.tracers
-    b = buoyancy_model.model
+    b_model = buoyancy_model.model
```
tomchor, line 146: This line doesn't read well. But I also don't have immediate suggestions on how to improve it... irrc the whole nomenclature of `BuoyancyModel` is kinda confusing and we discussed changing it over at Oceananigans a few times.

Maybe changing from `buoyancy_model` to `buoyancy_treatment` might help...?

test/runtests.jl
```diff
@@ -406,6 +410,68 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
...
+        b_sorted = b[p]
+
+        v_grid = VolumeField(model.grid)
+        A = sum(AreaField(model.grid))
+        z✶_check = -cumsum(reshape(v_grid, :)[p]) / A
+
+        sorted_buoyancy, z✶ = OneDReferenceField(model.tracers.b, rev = true)
+
+        @test isequal(reshape(interior(sorted_buoyancy), :), b_sorted)
+        @test isequal(reshape(interior(z✶), :), z✶_check)
```
tomchor, line 437: When comparing `b_sorted` and `sorted_buoiyancy`, it's always gonna return true by construction, right? Since `b_sorted` is calculated with the permutations that are used to calculate `sorted_buoyancy`? I think we should still keep this test (since it tests that the code is working and all), but maybe we need another one that's a little more rigorous.

tomchor: Would starting with linear buoyancy distribution in the `x` direction (where we know `db/dx`) and sorting it, making sure that `db/dz` matches, work? We could try that for grids that are stretched in the horizontal and vertical directions separately. If we could test something along those lines it would be great.

test/runtests.jl
```diff
@@ -406,6 +410,68 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
...
+        ρ_sorted = ρ_flat[p]
+
+        v_grid = VolumeField(model.grid)
+        A = sum(AreaField(model.grid))
+        z✶_check = -cumsum(reshape(v_grid, :)[p]) / A
+
+        sorted_density, z✶ = OneDReferenceField(ρ)
+
+        @test isequal(reshape(interior(sorted_density), :), ρ_sorted)
+        @test isequal(reshape(interior(z✶), :), z✶_check)
```
tomchor, line 471: The comment above extends to these other `model.buoyancy` types. A separate comment here is that you're writing a lot of code 3 times. You can probably reuse code, so that you write most things only once.

tomchor:
> Taking a step back @tomchor, @glwagner, @jbisits: Why do we actually care about the sorted buoyancy field or the "background state" as a function of physical space (i.e. as a 3D/4D array)? We never actually use that to calculate any terms in the BPE or APE budgets, as far as I can tell. All we ever need is z⋆(x,y,z,t) (as a 4D array) and dz⋆db|x,y,z,t (also as a 4D array), both of which are accurately and uniquely provided by the sorting methods we're discussing (aside from limit cases where the density/buoyancy of two grid cells are equal within machine precision, but that is a negligible effect). Since we have b(z⋆) and z⋆(b) as a sorted array, it is straight-forward to estimate the derivative dz⋆db and then map this back to each grid cell. There is never any need to average those fields and it doesn't make any sense to do so.
> 
> Am I missing something? (Feel free to suggest any additional calculations to make this discussion more concrete.)

I think indeed if we can get what we need without explicitly computing $b_\star$ that would be fine. I also think that once you do the hard work of computing the permutations (which if I understand correctly you need to get $z_\star$), then getting $b_\star$ is trivial anyway... right? @jbisits you're the one implementing this; what do you think?

tomchor:
> > Note that we still probably want to implement the expensive z⋆ calculation at some point since it's still needed for tilted domains or domains with immersed boundaries.
> 
> I don't think that's true. I've already implemented a sorting-based method for immersed boundaries (see my notebook linked above) and I think the method could be generalized to tilted domains but that one is more complicated because of the background buoyancy field and boundary conditions.

Ah, yes, I just now saw that comment. I'll look into it later.

> > Although at some point you mentioned that it can scale as nlog⁡(n)? I don't see how but hopefully you're right!
> 
> I meant that the sorting-based method scales like whatever `sort` scales like, which I think is nlog⁡(n). If we can get the sorting-based method to work for immersed boundaries then I don't see why we would want the brute-force n2 algorithm.

Agreed, if we can get the sorting method to work with immersed boundaries and tilted domains we don't need/want the brute-force approach of a double-loop heaviside iteration.

jbisits:
I am sorry I have been slow in replying to the recent comments. I will be able to get back to this around the middle of this coming week - my apologies!

jbisits:
Sorry to have fallen asleep on this PR! The way I was implementing this was to define a `OneDReferenceField` which has the total number of elements from the domain reshaped into a vector over the vertical extent of the domain,

```julia
function OneDReferenceField(f::Union{Field, AbstractOperation}; rev = false)

    area = sum(AreaField(f.grid))
    volume_field = VolumeField(f.grid)
    v = reshape(interior(volume_field), :)
    field_data = reshape(interior(f), :)

    p = sortperm(field_data; rev)
    sorted_field_data = field_data[p]
    z✶ = -cumsum(v[p]) / area # divide by area at surface

    grid_arch = f.grid.architecture
    grid_size = prod(size(f.grid))
    new_grid = RectilinearGrid(grid_arch, size = grid_size, z = (-f.grid.Lz, 0), topology=(Flat, Flat, Bounded))

    sorted_field = CenterField(new_grid)
    set!(sorted_field, reshape(sorted_field_data, size(new_grid)))
    z✶_field = CenterField(new_grid)
    set!(z✶_field, reshape(z✶, size(new_grid)))

    return sorted_field, z✶_field
end
```

It looks like [`CumulativeIntegral`](https://clima.github.io/OceananigansDocumentation/dev/appendix/library/#Oceananigans.AbstractOperations.CumulativeIntegral-Tuple{Oceananigans.Fields.AbstractField}) could simplify some of the calculation in what I have above. I don't really like the implementation above as there is a lot going on at each time step so I tried for some time to move this to a `KernelFunctionOperation`, or look for parts that could be simplified/optimized, but could not figure out how to and quite a bit of time went by very quickly (again my apologies!). I know there is quite a bit to be done for this calculation but returning new `field`s computed from modified model `field`s with new `grid`s even though it works seems like a lot of overhead.

Further, for this to work on grids that do not have uniform horizontal area I think the area at each depth of the water column still needs to be calculated (as [pointed out here](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2140214465)). With that in mind is there a way that @hdrake's method for non-uniform horizontal area to be put into a GPU kernel? I am happy to look into this if that is something that will give the best results.

@tomchor depending on what you think about the method above it may also be worth closing this PR and implementing any new changes/different methods in a new PR (maybe with this PR converted to an issue though have not looked into if that can be done). Or I can try to improve on what is already in place in this PR --- maybe the hiatus from this PR will have helped! I will leave everything as is until a decision is made.

Again sorry to have let this go!

tomchor:
@jbisits don't worry about it! We all have lots to do we have to prioritize :)

> It looks like [`CumulativeIntegral`](https://clima.github.io/OceananigansDocumentation/dev/appendix/library/#Oceananigans.AbstractOperations.CumulativeIntegral-Tuple%7BOceananigans.Fields.AbstractField%7D) could simplify some of the calculation in what I have above. I don't really like the implementation above as there is a lot going on at each time step so I tried for some time to move this to a `KernelFunctionOperation`, or look for parts that could be simplified/optimized, but could not figure out how to and quite a bit of time went by very quickly (again my apologies!). I know there is quite a bit to be done for this calculation but returning new `field`s computed from modified model `field`s with new `grid`s even though it works seems like a lot of overhead.

I don't think this can be implemented using `KernelFunctionOperation`s, since you're rearranging the grid. But I'd be happy to be proven wrong.

> Further, for this to work on grids that do not have uniform horizontal area I think the area at each depth of the water column still needs to be calculated (as [pointed out here](https://github.com/tomchor/Oceanostics.jl/pull/172#issuecomment-2140214465)). With that in mind is there a way that @hdrake's method for non-uniform horizontal area to be put into a GPU kernel? I am happy to look into this if that is something that will give the best results.
>
> @tomchor depending on what you think about the method above it may also be worth closing this PR and implementing any new changes/different methods in a new PR (maybe with this PR converted to an issue though have not looked into if that can be done). Or I can try to improve on what is already in place in this PR --- maybe the hiatus from this PR will have helped! I will leave everything as is until a decision is made.
> 
> Again sorry to have let this go!

It does look like @hdrake's method would give the best results. I haven't looked at that method in a while (so I'll let @hdrake chime in here) but if this PR has something that already _works_ (even though it's limited to regular grids and you're not 100% happy with the main function) I'd say it's okay to merge. Then we can improve this design later or even replace it if we think it's necessary. Personally, I think having something you can already use helps to visualize better ways of doing things.

tomchor:
@jbisits let me know if you still would like to finish this PR and I'll resolve the merge conflicts. Basically I re-organized all the tests for make things faster and more comprehensive, but it should affect what you already did.

jbisits:
> @jbisits let me know if you still would like to finish this PR and I'll resolve the merge conflicts. Basically I re-organized all the tests for make things faster and more comprehensive, but it should affect what you already did.

Hi @tomchor, I would like to get the APE diagnostic but I think there are better ways to do it. I have been experimenting locally with some of these. Not sure what your thoughts are but given the staleness of this PR I would be in favour of opening a new PR for this diagnostic (I have started one on a branch). Let me know your thoughts because I know you said above it might be better to merge and then fix but I think I can improve on what I tried here.

tomchor:
> > @jbisits let me know if you still would like to finish this PR and I'll resolve the merge conflicts. Basically I re-organized all the tests for make things faster and more comprehensive, but it should affect what you already did.
> 
> Hi @tomchor, I would like to get the APE diagnostic but I think there are better ways to do it. I have been experimenting locally with some of these. Not sure what your thoughts are but given the staleness of this PR I would be in favour of opening a new PR for this diagnostic (I have started one on a branch). Let me know your thoughts because I know you said above it might be better to merge and then fix but I think I can improve on what I tried here.

They conflicts are pretty easy to resolve and I can do that for you. So I wouldn't make any decisions based on the "staleness" of the PR.

That said, if you think it's better to close this PR and start fresh with a better implementation, then I'm all for that. I haven't thought about this in a while, so I'll defer to you in that regard.

jbisits:
> That said, if you think it's better to close this PR and start fresh with a better implementation, then I'm all for that. I haven't thought about this in a while, so I'll defer to you in that regard.

Thanks @tomchor! I am about to start working on BPE/APE related things again in the next week so I will have this at the front of my mind. I will try to get what I have been working on locally done in the next month. I will leave this open for now and be in touch! Thanks again for your patience with me getting this done.

tomchor:
> > That said, if you think it's better to close this PR and start fresh with a better implementation, then I'm all for that. I haven't thought about this in a while, so I'll defer to you in that regard.
> 
> Thanks @tomchor! I am about to start working on BPE/APE related things again in the next week so I will have this at the front of my mind. I will try to get what I have been working on locally done in the next month. I will leave this open for now and be in touch! Thanks again for your patience with me getting this done.

For sure! Thanks for the contribution!

tomchor:
Superseded by https://github.com/tomchor/Oceanostics.jl/pull/272

PR #176, opened 2024-05-20
Correct sign inconsistency in `PotentialEnergy` diagnostics

### LINKED ISSUE #175: Sign inconsistency in `PotentialEnergy` diagnostics?
If buoyancy is defined by $b = -g\frac{\rho}{\rho_{0}}$, then shouldn't potential energy density be $E_{p} = - zb$ instead of the currently implemented $E_{p} = zb$? Is there an implicit negative sign somewhere that I am missing?

https://github.com/tomchor/Oceanostics.jl/blob/d2060579e175cd4ac3265a165cfd49d2ac304355/src/PotentialEnergyEquationTerms.jl#L164-L187

jbisits:
Thanks @hdrake for finding this! The changes are `bz_ccc` -> `minus_bz_ccc` plus defines $E_{p} = \frac{g \rho}{\rho_{0}}z = -bz$ in the docstring.

Closes #175

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3,6 +3,7 @@ using CUDA
 
 using Oceananigans
 using Oceananigans.AbstractOperations: AbstractOperation
+using Oceananigans.BuoyancyModels: buoyancy_perturbationᶜᶜᶜ
 using Oceananigans.Fields: @compute
 using Oceananigans.TurbulenceClosures: ThreeDimensionalFormulation
 using Oceananigans.Models: seawater_density, model_geopotential_height
@@ -407,6 +408,27 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
 
     return nothing
 end
+function test_PEbuoyancytracer_PElineareos(grid)
+
+    model_buoyancytracer = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=:b)
+    model_lineareos = NonhydrostaticModel(; grid, buoyancy=SeawaterBuoyancy(), tracers=(:S, :T))
+    C_grad(x, y, z) = 0.01 * z
+    set!(model_lineareos, S = C_grad, T = C_grad)
+    linear_eos_buoyancy(grid, buoyancy, tracers) =
+        KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, buoyancy, tracers)
+    b_field = Field(linear_eos_buoyancy(model_lineareos.grid, model_lineareos.buoyancy.model, model_lineareos.tracers))
+    compute!(b_field)
+    set!(model_buoyancytracer, b = interior(b_field))
+    pe_buoyancytracer = Field(PotentialEnergy(model_buoyancytracer))
+    compute!(pe_buoyancytracer)
+    pe_lineareos = Field(PotentialEnergy(model_lineareos))
+    compute!(pe_lineareos)
+
+    @test all(interior(pe_buoyancytracer) .== interior(pe_lineareos))
+
+    return nothing
+
+end
 #---
 
 #+++ Known-value function tests
@@ -596,6 +618,7 @@ model_types = (NonhydrostaticModel, HydrostaticFreeSurfaceModel)
                 end
 
             end
+            test_PEbuoyancytracer_PElineareos(grid)
 
         end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -618,7 +618,7 @@ model_types = (NonhydrostaticModel, HydrostaticFreeSurfaceModel)
                 end
 
             end
-            test_PEbuoyancytracer_PElineareos(grid)
+            test_PEbuoyancytracer_equals_PElineareos(grid)
 
         end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -408,7 +408,7 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
 
     return nothing
 end
-function test_PEbuoyancytracer_PElineareos(grid)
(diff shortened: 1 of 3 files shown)


tomchor:
Thanks @jbisits, and thanks @hdrake for catching this!

Two suggestions:

1. I think we should add a test that would catch this in the future. Maybe just ensuring that all versions of `PotentialEnergy` (with `BuoyancyTracer`, `Seawater`, etc...) produce the same result for equivalent profiles would be enough. Thoughts?
2. Can you bump the patch version before merging? We should register a bugfix release right afterwards.

jbisits:
> I think we should add a test that would catch this in the future. Maybe just ensuring that all versions of PotentialEnergy (with BuoyancyTracer, Seawater, etc...) produce the same result for equivalent profiles would be enough. Thoughts?

I think something like this is a good idea and will be useful to have to check `BackgroundPotentialEnergy` (when it is done..) produces equivalent results.

jbisits:
I added a test to compare the `PotentialEnergy` computation in the `BuoyancyTracer` and `LinearEquationOfState` cases, @tomchor let me know if this is the type of thing you were thinking of.

I am not quite sure how to test against the other method as for `BoussinesqEquationOfState` there are several options for the equation of state. There is already a test that for the computation of `PotentialEnergy` when there is a `BoussinesqEquationOfState` that performs a simple check which may be sufficient. Let me know what you think.

test/runtests.jl
```diff
@@ -407,6 +408,27 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
 
     return nothing
 end
+function test_PEbuoyancytracer_PElineareos(grid)
```
tomchor, line 411:
```suggestion
function test_PEbuoyancytracer_equals_PElineareos(grid)
```

test/runtests.jl
```diff
@@ -407,6 +408,27 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
...
+
+    model_buoyancytracer = NonhydrostaticModel(; grid, buoyancy=BuoyancyTracer(), tracers=:b)
+    model_lineareos = NonhydrostaticModel(; grid, buoyancy=SeawaterBuoyancy(), tracers=(:S, :T))
+    C_grad(x, y, z) = 0.01 * z
+    set!(model_lineareos, S = C_grad, T = C_grad)
+    linear_eos_buoyancy(grid, buoyancy, tracers) =
+        KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, buoyancy, tracers)
+    b_field = Field(linear_eos_buoyancy(model_lineareos.grid, model_lineareos.buoyancy.model, model_lineareos.tracers))
+    compute!(b_field)
+    set!(model_buoyancytracer, b = interior(b_field))
```
tomchor, line 421: Clever!

test/runtests.jl
```diff
@@ -407,6 +408,27 @@ function test_potential_energy_equation_terms(model; geopotential_height = nothi
...
+    pe_buoyancytracer = Field(PotentialEnergy(model_buoyancytracer))
+    compute!(pe_buoyancytracer)
+    pe_lineareos = Field(PotentialEnergy(model_lineareos))
+    compute!(pe_lineareos)
+
+    @test all(interior(pe_buoyancytracer) .== interior(pe_lineareos))
```
tomchor, line 427: I guess we're only testing CPU architecture in this function, but I think this will fail on GPUs. I'm okay leaving it like this since I think `PotentialEnergy()` is tested in a way that's compatible with GPUs somewhere else. It is, right?

tomchor: Relatedly, I'm a little out of the loop about buoyancy models. This is all the models we currently have available, correct?

https://github.com/tomchor/Oceanostics.jl/blob/d2060579e175cd4ac3265a165cfd49d2ac304355/test/runtests.jl#L526-L528

test/runtests.jl
```diff
@@ -596,6 +618,7 @@ model_types = (NonhydrostaticModel, HydrostaticFreeSurfaceModel)
                 end
 
             end
+            test_PEbuoyancytracer_PElineareos(grid)
```
tomchor, line 621: Same change needs to be made here

```suggestion
            test_PEbuoyancytracer_equals_PElineareos(grid)
```

test/runtests.jl
jbisits: Yes I think this all buoyancy models but it does not test all the [second order `RoquetEquationOfState`s](https://clima.github.io/SeawaterPolynomials.jl/dev/API/#SeawaterPolynomials.SecondOrderSeawaterPolynomials.RoquetSeawaterPolynomial) which we can do but I thought two would be enough.

jbisits: > I guess we're only testing CPU architecture in this function, but I think this will fail on GPUs. I'm okay leaving it like this since I think `PotentialEnergy()` is tested in a way that's compatible with GPUs somewhere else. It is, right?

Yes the function setup and computation tested is tested here 
https://github.com/tomchor/Oceanostics.jl/blob/d2060579e175cd4ac3265a165cfd49d2ac304355/test/runtests.jl#L387-L390
but it only checks that it can set the `Field` and can then `compute!` the `Field` i.e. not testing it against another value.
I will try and test on a `GPU()` today.

jbisits: > I guess we're only testing CPU architecture in this function, but I think this will fail on GPUs. I'm okay leaving it like this since I think `PotentialEnergy()` is tested in a way that's compatible with GPUs somewhere else. It is, right?

This test passed when I ran it on the GPU, so I think we should be ok!

tomchor (approved):
LGTM! Feel free to both merge and register a new version after that if you're comfortable doing it.

Issue #159, opened 2023-12-28
Animation in tilted bottom boundary layer example doesn't work in the docs when the grid is stretched


tomchor:
This is a problem that popped up in https://github.com/tomchor/Oceanostics.jl/pull/157 and I haven't been able to fix it.

Basically when the grid there is stretched, the docs hang at the point where the animation is being created.
- This does not happen when running the example locally (i.e. everything works without issues)
- And this issue does not happen with a regular grid.

So for now https://github.com/tomchor/Oceanostics.jl/pull/157 re-wrote the example using a regular grid and things seems to be working properly. I believe the problem is either in the interaction between Documenter and Makie or in the interaction between Documenter and Rasters.

What makes it hard to investigate this is that I wasn't able to reproduce it locally at all :/
So for each iteration I've been having to make a change, commit, push it to a PR, wait for about an hour for the docs to get there and see the CI hangs. I think the next step is to make a MWE with an animation without Rasters, to at least narrow down if the issue is with Makie + Documenter, or Rasters + Documenter.

glwagner:
Maybe try speeding up the docs build in CI by commenting everything out except this one example in `make.jl`.

tomchor:
I've done that in the course of #157, but it's slow since there's no error, so it forces me to wait for a longer-than-usual amount of time to see if indeed nothing is happening.

glwagner:
You mean even with only one example being loaded the docs still take an hour to run? 😱

glwagner:
Can you paste the code that creates the animation?

tomchor:
Here!:

https://github.com/tomchor/Oceanostics.jl/blob/2036fe3780efb6e8a8b7d386e506da446be976bf/docs/examples/tilted_bottom_boundary_layer.jl#L167-L218

glwagner:
Huh, I'm not familiar with the packages being used. It seems probable that the relative slowdown for stretched vs uniform vertical grid has to do with `:zF => Z`. What's `Z`?

PR #177, opened 2024-06-19
Add stretched coordinates back to tilted BBL example

tomchor:
Stretched coordinates removed from the tilted bottom boundary layer example in https://github.com/tomchor/Oceanostics.jl/pull/157 because something about it was causing the docs to get stuck. This PR stretches the vertical coordinate on that grid again.

Closes https://github.com/tomchor/Oceanostics.jl/issues/159

### DIFF
--- a/docs/examples/tilted_bottom_boundary_layer.jl
+++ b/docs/examples/tilted_bottom_boundary_layer.jl
@@ -201,7 +201,7 @@ xF = Array(dims(ds, :xF))
 zC = Array(dims(ds, :zC))
 zF = Array(dims(ds, :zF))
 
-bₙ = @lift set(ds.b[Ti=$n, yC=Near(0)], :xC => X, :zC => Z)
+bₙ = @lift Array(ds.b[Ti=$n, yC=Near(0)])
 
 Riₙ = @lift Array(ds.Ri[Ti=$n, yC=Near(0)])
 hm1 = heatmap!(ax1, xC, zF, Riₙ; colormap = :coolwarm, colorrange = (-1, +1))
--- a/docs/examples/tilted_bottom_boundary_layer.jl
+++ b/docs/examples/tilted_bottom_boundary_layer.jl
@@ -17,7 +17,7 @@
 #
 # ## Grid
 #
-# We start by creating a ``x, z`` grid with 64² cells:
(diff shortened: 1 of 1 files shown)


glwagner:
Nice.

Happy to merge this but curious whether this matters much. Stretching often seems like a complication that _feels_ more sophisticated, but doesn't really give much benefit (at least for wall-modeled boundary layer turbulence problems where usually at least half the domain is fully turbulent and we don't need super fine resolution near the boundary).

glwagner (approved):

tomchor:
I agree that it's not very important. The main purpose of the example is to show the `ProgressMessenger` and diagnostics capabilities and it does that with a regular just fine.

That said, I do think stretching makes the symmetric/centrifugal instabilities way more resolved at the bottom, with our small amount of points. Compare the level of noise in the PV or Ro number in the videos generated by [current docs](https://tomchor.github.io/Oceanostics.jl/v0.14.2/generated/tilted_bottom_boundary_layer/) versus the [old docs with stretched coords](https://tomchor.github.io/Oceanostics.jl/v0.13.1/generated/tilted_bottom_boundary_layer/). So I do think it's a positive change, although not a priority. So if I can't make it happen relatively easily with this PR I'll probably drop the idea.

tomchor:
Also an issue I've just noticed. It seems that something changed in the way the diagnostics are calculated and now the PV has an artifact at the west wall even though the domain is periodic in `x`:

image

This seems to have started with [version 0.14.1](https://tomchor.github.io/Oceanostics.jl/v0.14.1/generated/tilted_bottom_boundary_layer/) and olders versions don't have that.

Looking at the changelog between 0.14.0 and 0.14.1, I don't anything there that could have caused this: https://github.com/tomchor/Oceanostics.jl/compare/v0.14.0...v0.14.1

It was during that version that we switched to Julia 1.10 and Oceananigans jumped to version 0.91, so maybe something  in Oceananigans changed?

glwagner:
You're missing a fill halo regions somewhere perhaps?

glwagner:
You may want to manually set the number of halos. You need 2 halo points? Or 3?

Probably you want to break down the PV term by term to see which term exhibits issues and then diagnose that.

PR #182, opened 2024-11-11
Adds a mixed layer depth diagnostic

jagoosw:
Hi @tomchor,

This PR adds a simple mixed layer depth diagnostic using a simple density anomaly criteria.

I don't think there is a convenient way to generally write a mixed layer depth computation, so I have called it `DensityCriteriaMixedLayerDepth`.


tomchor:
Nice! I'm okay with the name, although  I think it should be density _criterion_, and not criteria (single it's one, singular, criterion, no?).

Another alternative is to create an object called `MixedLayerDepth` that takes a criterion (or algorithm to be more precise) as the computation. That would make it easier in the future to have other way of calculating the MLD. For now the only algorithm would be the density threshold criterion.

Also, I see you're using a `KernelFunctionOperation`, but you want to return a 2D object, no? Is this the best way to do things? (I'm genuinely asking; I'm not sure.)

tomchor:
(Also github asked for explicit approval to run the tests and docs. I have no idea why, but we should figure out how to disable that!)

jagoosw:
> Nice! I'm okay with the name, although I think it should be density _criterion_, and not criteria (single it's one, singular, criterion, no?).

Ah yes you are right, I have changed now.

> Another alternative is to create an object called `MixedLayerDepth` that takes a criterion (or algorithm to be more precise) as the computation. That would make it easier in the future to have other way of calculating the MLD. For now the only algorithm would be the density threshold criterion.

I've done this now, I'm not sure which is better but you can have a look and I can roll back the commit if you prefer the first way.

> Also, I see you're using a `KernelFunctionOperation`, but you want to return a 2D object, no? Is this the best way to do things? (I'm genuinely asking; I'm not sure.)

This is a good question, it seems to work for 2D objects by specifying a nothing location like `KernelFunctionOperation{Center, Center, Nothing}` and you can index with just two indices like `mld[i, j]`

jagoosw:
> (Also github asked for explicit approval to run the tests and docs. I have no idea why, but we should figure out how to disable that!)

I think this is the default behaviour for first-time contributors to a repo incase the CI has some kind of cost associated with it

tomchor:
@jagoosw Sorry, I realize I didn't check this PR after your latest pushes. Conflicts aside (I can resolve them for you), is this PR close/ready to merge?

jagoosw:
No problem, I forgot all about this too! I think its all ready if you're happy with it apart from the conflicts

tomchor:
> No problem, I forgot all about this too! I think its all ready if you're happy with it apart from the conflicts

Nice! Good to hear. I'll work on the conflicts today and take a closer look. If all is good, I'll approve and then you're free to merge. I invited you to become a collaborator here if you want, that way you can merge your own PRs.

test/runtests.jl
```diff
@@ -23,6 +45,10 @@ group     = get(ENV, "TEST_GROUP", :all) |> Symbol
         include("test_canonical_flows.jl")
     end
 
+    test_PEbuoyancytracer_equals_PElineareos(grid)
+    test_mixed_layer_depth(grid)
+
+
```
tomchor, line 51:
```suggestion
    test_mixed_layer_depth(grid)
```

tomchor:
Ah, I just realized you're using your own fork to implement this so I can't modify it!

I think all you need to do is to move the test from `runtests.jl` to `test_tracer_diagnostics.jl` for now and then we can go from there.

src/FlowDiagnostics.jl
```diff
@@ -4,18 +4,24 @@ using DocStringExtensions
...
 export StrainRateTensorModulus, VorticityTensorModulus, Q, QVelocityGradientTensorInvariant
+export MixedLayerDepth, DensityAnomalyCriterion
 
 using Oceanostics: validate_location,
                    validate_dissipative_closure,
                    add_background_fields,
                    get_coriolis_frequency_components
 
 using Oceananigans: NonhydrostaticModel, FPlane, ConstantCartesianCoriolis, BuoyancyField, BuoyancyTracer
+using Oceananigans.BuoyancyModels: get_temperature_and_salinity, SeawaterBuoyancy
```
tomchor, line 15:
```suggestion
using Oceananigans.BuoyancyFormulations: get_temperature_and_salinity, SeawaterBuoyancy
```

test/runtests.jl
```diff
@@ -2,6 +2,28 @@ using Test
 
 group     = get(ENV, "TEST_GROUP", :all) |> Symbol
 
+function test_mixed_layer_depth(grid;
```
tomchor, line 5:
```suggestion
function test_mixed_layer_depth(grid;
```

test/runtests.jl
```diff
@@ -2,6 +2,28 @@ using Test
...
+    mld = MixedLayerDepth(grid, buoyancy, (; T))
+
+    @test isinf(mld[1, 1])
+    
+    set!(T, (x, y, z) -> z * ∂z_T+10)
+    fill_halo_regions!(T)
+
+    @test mld[1, 1] ≈ -zₘₓₗ
+
+    return nothing
```
tomchor, line 24:
```suggestion
                                                            constant_salinity=0),
                                zₘₓₗ = 0.5,
                                δρ = 0.125,
                                ∂z_T = - δρ / zₘₓₗ / buoyancy.equation_of_state.seawater_polynomial.R₀₁₀)
    boundary_conditions = FieldBoundaryConditions(grid, (Center, Center, Center); top = GradientBoundaryCondition(∂z_T))

    T = CenterField(grid; boundary_conditions)
    mld = MixedLayerDepth(grid, buoyancy, (; T))
    @test isinf(mld[1, 1])

    set!(T, (x, y, z) -> z * ∂z_T+10)
    fill_halo_regions!(T)

    @test mld[1, 1] ≈ -zₘₓₗ

    return nothing
```

tomchor:
@jagoosw if you'd like me to fix things you can make me a collaborator on your fork I guess? Not sure if there's a better way since this PR is already created based on it.

glwagner:
You should not use a density criteria as it is not general. A buoyancy criteria is better.

tomchor:
> You should not use a density criteria as it is not general. A buoyancy criteria is better.

The way I understand the code, different criteria can be plugged into `MixedLayerDepth`, no? Having a density criterion doesn't preclude having a buoyancy criterion.

That said, can you elaborate why density criteria are less general than buoyancy criteria?

glwagner:
> That said, can you elaborate why density criteria are less general than buoyancy criteria?

Gravitational forces enters a Boussinesq model via buoyancy. One can simulate buoyancy directly, leading to the formation of "mixed layers". But density is not defined in such a model.

More generally, typical mixed layer depth criterion actually use some potential density rather than in-situ density, because they do not want to incorporate the dynamically irrelevant part of density due to compressibility.

Using buoyancy is simpler and avoids those issues; by design, the buoyancy field can be computed from any model that has gravitational forces.

It's nice to have different criteria but I would discourage developing any criteria other than ones based on the buoyancy profile.

tomchor:
@jagoosw I moved the test to its proper place, but it's failing with some version of

```
Tracer diagnostics tests: Test Failed at /home/tomas/repos/Oceanostics.jl/test/test_tracer_diagnostics.jl:194
  Expression: mld[1, 1] ≈ -zₘₓₗ
   Evaluated: 1.3877787807814457e-17 ≈ -0.5
```

Can you please take a look?

jagoosw:
I've added a buoyancy criterion now, and it makes sense that we should prefer that since it will work in all models, whereas the density one only worked with `BoussinesqEquationOfState`.

The test also passes locally for me now, but I have no idea what made it work again.

src/FlowDiagnostics.jl
```diff
@@ -486,56 +486,89 @@ function (MLD::MixedLayerDepth)(i, j, k, grid, args...)
...
 
-validate_criterion_model(::DensityAnomalyCriterion, buoyancy, C) = 
-    @error "DensityAnomalyCriterion is not implemented for your buoyancy, model as the density anomaly is not defined"
+validate_criterion_model(::DensityAnomalyCriterion, buoyancy, C) = nothing
     
 validate_criterion_model(::DensityAnomalyCriterion, ::SeawaterBuoyancy{<:Any, <:BoussinesqEquationOfState}, C::NamedTuple) = nothing
 
-@inline function density(i, j, k, grid, b::SeawaterBuoyancy{<:Any, <:BoussinesqEquationOfState}, C)
+@inline function anomaly(::DensityAnomalyCriterion, i, j, k, grid, b::SeawaterBuoyancy{<:Any, <:BoussinesqEquationOfState}, C)
     T, S = get_temperature_and_salinity(b, C)
```
glwagner, line 548:
```suggestion
```

src/FlowDiagnostics.jl
```diff
@@ -486,56 +486,89 @@ function (MLD::MixedLayerDepth)(i, j, k, grid, args...)
...
+greater than the surface buoyancy (but the pertubaton is usually negative).
 
-    return ρₖ > ρᵣ + δρ
+When this model is used, the arguments `buoyancy` and `C` should be 
+supplied where `buoyancy` should be the buoyancy model, and `C` should be a named 
+tuple of `(; T, S)`, `(; T)` or `(; S)` (the latter two if the buoyancy model 
+specifies a constant salinity or temperature).
+"""
+@kwdef struct BuoyancyAnomalyCriterion{FT} <: AbstractAnomalyCriterion
+    pertubation :: FT = - 1/8 * 1020 / g_Earth
```
glwagner, line 564: Something is off here. Buoyancy has units of gravitational acceleration (not 1/g). So it makes sense to express the buoyancy anomaly as `c * g`, where `c` is some non-dimensional number that represents your interpretation of a reasonable mixed layer depth for Boussinesq.

src/FlowDiagnostics.jl
```diff
@@ -486,56 +486,89 @@ function (MLD::MixedLayerDepth)(i, j, k, grid, args...)
...
 When this model is used, the arguments `buoyancy` and `C` should be 
-supploed where `buoyancy` should be the buoyancy model, and `C` should be a named 
+supplied where `buoyancy` should be the buoyancy model, and `C` should be a named 
 tuple of `(; T, S)`, `(; T)` or `(; S)` (the latter two if the buoyancy model 
 specifies a constant salinity or temperature).
 """
-struct DensityAnomalyCriterion{FT}
+struct DensityAnomalyCriterion{FT} <: AbstractAnomalyCriterion
     pertubation :: FT
 end
```
glwagner, line 537: The density anomaly can be expressed in terms of a buoyancy anomaly, where you require additional information about 1) the gravitational acceleration used in the model and 2) the reference density. For example:

```julia
struct DensityAnomalyCriterion
    buoyancy_anomaly
    reference_density
    gravitational_acceleration
end
```

You can then build a constructor that uses some default values for the parameters.

This will help users understand how the parameters they are using affect their results leading to higher-quality science.

src/FlowDiagnostics.jl
jagoosw: Oh I put this the wrong way around it should be - g * 0.125 / 1020

glwagner: How about just `- 1e-4 * g` as the default?

src/FlowDiagnostics.jl
jagoosw: I like this more

tomchor:
> The test also passes locally for me now, but I have no idea what made it work again.

It seems like the errors are still there! I wonder why you can't reproduce these errors locally.

jagoosw:
Ah its because the stretched grid doesn't have z=0 at the top

src/FlowDiagnostics.jl
```diff
@@ -441,6 +446,137 @@ function QVelocityGradientTensorInvariant(model; location = (Center, Center, Cen
...
+tuple of `(; T, S)`, `(; T)` or `(; S)` (the latter two if the buoyancy model 
+specifies a constant salinity or temperature).
+"""
+function MixedLayerDepth(grid, args...; criterion = BuoyancyAnomalyCriterion(convert(eltype(grid), -1e-4 * g_Earth)))
+    validate_criterion_model(criterion, args...)
+
+    MLD = MixedLayerDepth(criterion)
+
+    return KernelFunctionOperation{Center, Center, Nothing}(MLD, grid, args...)
+end
```
tomchor, line 473: I think the intention here is for the method to be called when `MLD = MixedLayerDepth(grid)`, right?

This isn't happening right now:
```julia
julia> @which Oceanostics.MixedLayerDepth(grid)
MixedLayerDepth(criterion::C) where C
     @ Oceanostics.FlowDiagnostics ~/repos/Oceanostics.jl2/src/FlowDiagnostics.jl:451
```
 Can confirm? If I'm right, the signature probably should be:

```suggestion
function MixedLayerDepth(grid::AbstractGrid, args...; criterion = BuoyancyAnomalyCriterion(convert(eltype(grid), -1e-4 * g_Earth)))
    validate_criterion_model(criterion, args...)

    MLD = MixedLayerDepth(criterion)

    return KernelFunctionOperation{Center, Center, Nothing}(MLD, grid, args...)
end
```

tomchor:
I like the science part here. However, I'm a little confused by the implementation. `MixedLayerDepth` is defined as a a struct which takes only a `criterion`. But then if one calls something like `MixedLayerDepth(grid, buoyancy, C; criterion)`, then the following method is called:
https://github.com/tomchor/Oceanostics.jl/blob/f9ac0e10f522b3c577102b17b99352944832ab4a/src/FlowDiagnostics.jl#L467-L473

Which actually returns a `KernelFunctionOperation`. Am I correct?

If so, that feels a bit confusing/convoluted to me. Is there a way to make this more straightforward?

@glwagner I think this overlaps with our discussion in https://github.com/tomchor/Oceanostics.jl/pull/194, so maybe some points are still not connecting in my head...

src/FlowDiagnostics.jl
glwagner: This is also incorrect because it doesn't return `MixedLayerDepth`. I suggest defining an alias called `MixedLayerDepthOperation` and then using that as the name for the constructor. You could also change `MixedLayerDepth` to something else (like `MixedLayerDepthKernel`) and reserv ethe name `MixedLayerDepth` for the operation.

tomchor:
@jagoosw I addressed both mine and @glwagner's comments directly. Let me know if you don't agree with anything I'll revert the commits.

tomchor (approved):

PR #190, opened 2024-12-18
Generalize `ErtelPotentialVorticity` and `RichardsonNumber` to work when `model.buoyancy` is not `BuoyancyTracer`

tomchor:
Easy fix. I wasn't aware that the function `Oceananigans.BuoyancyFormulations.buoyancy` existed when I first coded EPV.

### DIFF
--- a/src/FlowDiagnostics.jl
+++ b/src/FlowDiagnostics.jl
@@ -187,17 +187,17 @@ is defined as
 where `f` is the Coriolis frequency, `ωᶻ` is the relative vorticity in the `z` direction, `b` is the buoyancy, and
 `∂U/∂z` and `∂V/∂z` comprise the thermal wind shear.
 """
-function ThermalWindPotentialVorticity(model; location = (Face, Face, Face))
+function ThermalWindPotentialVorticity(model; tracer=:b, location = (Face, Face, Face))
     validate_location(location, "ThermalWindPotentialVorticity", (Face, Face, Face))
     u, v, w = model.velocities
-    return ThermalWindPotentialVorticity(model, u, v, buoyancy(model), model.coriolis; location)
+    return ThermalWindPotentialVorticity(model, u, v, model.tracers[tracer], model.coriolis; location)
 end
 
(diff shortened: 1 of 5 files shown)


src/FlowDiagnostics.jl
```diff
@@ -279,25 +280,16 @@ Note that EPV values are correctly calculated both in the interior and the bound
...
 is zero there.
 """
-function ErtelPotentialVorticity(model; location = (Face, Face, Face), add_background = true)
+function ErtelPotentialVorticity(model; location = (Face, Face, Face))
     validate_location(location, "ErtelPotentialVorticity", (Face, Face, Face))
 
-    if model.buoyancy == nothing || !(model.buoyancy.formulation isa BuoyancyTracer)
-        throw(ArgumentError("`ErtelPotentialVorticity` is only implemented for `BuoyancyTracer` at the moment."))
+    if model.buoyancy == nothing
+        throw(ArgumentError("Cannot calculate `ErtelPotentialVorticity` when `model.buoyancy == nothing`."))
```
glwagner, line 287: You can actually use any tracer to compute PV (buoyancy just gives nice results because it's 1-1 with z coordinate).

glwagner: so you could let people specify the tracer which defaults to buoyancy

tomchor: Ah that's true, I forgot about that detail. I guess I don't need to transform active tracers into buoyancy...

glwagner: Buoyancy is usually the best choice so I do think it's a good idea to use it. I'm just saying without any additional work you could allow users to chose the "stretching tracer". The idea is that $\Pi_c = \bf{\omega}^a \cdot \bnabla c$ is conserved whenever $c$ is conserved.

tomchor: Yeah, 100% agree. I simply forgot that the PV definition is more general than what we end up using in practice!

Issue #196, opened 2025-03-19
Tests are taking too long


tomchor:
It's not a major issue at the moment, but test are currently taking about 2.5 hours. See for example https://github.com/tomchor/Oceanostics.jl/pull/195

This is because tests in Oceanostics have nested loops that test pretty much every combination of grids, closures, etc. In a way that's good because it covers a huge number of cases and combinations, but on the other hand it might be unnecessarily thorough, and as the package grows and tests multiply, this becomes exponentially slow.

glwagner:
I'm not sure what this does:

https://github.com/tomchor/Oceanostics.jl/blob/875f330e0db81068b4cffce524bd5388507cc99f/.github/workflows/ci.yml#L26

but one way you can reduce test time is to run with optimization level `-O0`. Testing is probably 99% compile time, and O0 can cut compile time in half.

I also agree that it's nice but not necessary to test everything. We should test common and representative cases, and add edge cases if we anticipate or find a problem.

tomchor:
> I'm not sure what this does:
> 
> [Oceanostics.jl/.github/workflows/ci.yml](https://github.com/tomchor/Oceanostics.jl/blob/875f330e0db81068b4cffce524bd5388507cc99f/.github/workflows/ci.yml#L26)
> 
> Line 26 in [875f330](/tomchor/Oceanostics.jl/commit/875f330e0db81068b4cffce524bd5388507cc99f)
>  - uses: julia-actions/julia-runtest@v1 

It's what actually runs the tests: https://github.com/julia-actions/julia-runtest. Not sure how to change the compile flag there though...

glwagner:
Sorry yeah I gathered it ran the tests, but I don't know what command it uses. However, you can basically just write 

```
julia --project -O0 -e 'using Pkg; Pkg.test()'
```

if you want to configure things manually. Or maybe there is a way to pass options to that utility

glwagner:
You can also break up the test into different actions

tomchor:
> You can also break up the test into different actions

Yeah that's a good point! I'll still investigate how to pass an optimization flag to julia though.

tomchor:
Two things:

1. Currently there's no way to pass optimization flag to `julia-runtest`, but there's a PR to implement that: https://github.com/julia-actions/julia-runtest/pull/140
[...]

PR #197, opened 2025-03-27
Separate tests

tomchor:
Helps with https://github.com/tomchor/Oceanostics.jl/issues/196#issuecomment-2739072025

cc @glwagner


tomchor:
Down from 2.5h to 1.5h. I think it's good enough to close this PR, and I'll consider other ways of speeding things up in future PRs.

PR #203, opened 2025-05-06
Adds `BottomCellValue`

tomchor:
This PR adds `BottomCellValue`, which gets the value of a given `diagnostic` (or `Field` or `AbstractOp` at the bottom, which may be the bottom of the domain or an immersed boundary bottom. At the moment this is programmed very simply by creating the `KernelFunctionOperation` ~called `bottom_adjacent_node()`~ (which checks if a given cell/node is immediately above a bottom) and using that as a mask to multiple whatever you want to get the bottom value of.


glwagner:
I think this may more properly belong in Oceananigans

glwagner:
what I mean is that this would be the building block of such a feature (so we need `BottomCellValue` in order to build `MappedKernelOperation`)

PR #204, opened 2025-06-02
Reorganize module and notations

tomchor:
Closes https://github.com/tomchor/Oceanostics.jl/issues/35 https://github.com/tomchor/Oceanostics.jl/issues/138 https://github.com/tomchor/Oceanostics.jl/issues/207

This PR will reorganize Oceanostics, as well as revamp the naming conventions (accordingly changing functions from camel-case).


tomchor:
@glwagner I'm wondering if I can get your opinion here. At the moment this PR is re-working the (most) modules to be named `TracerEquation`, `TracerVarianceEquation`, etc. And then within that, we'd have each term of the equation. I'd like to not export each term at the package level, only at the module level. So for example, if you want the tracer variance dissipation rate, you'd have to call:

```julia
using Oceanostics
D = TracerVarianceEquation.DissipationRate(model)
```
And if you just did
```julia
using Oceanostics
D = DissipationRate(model)
```
this wouldn't work.

I think that works, but has two disadvantages:

1. If a user did `using Oceanostics.TracerVarianceEquation: DissipationRate`, then suddenly `DissipationRate` is ambiguous by itself (and can conflict with other dissipation rates from other modules)
2. When we want to calculate the variable which gives name to the equation, we'd be forced to do `TracerVarianceEquation.TracerVariance`, which is kinda awkward.

We could, of course, define all types (which correspond to equation terms) with their full name and export everything, which is kind of what Oceanostics has been doing. So for example, we'd have:

```julia
using Oceanostics.TracerVarianceEquation: TracerVarianceDissipationRate
```
which is kind of what we were doing before this PR, but I find it a bit unorganized (plus the names can become quite long).

Curious about your thoughts.

tomchor:
Maybe one third option would be to create a different abstract type such as `AbstractBudgetedQuantity`. And then create structs for each quantity of interest that would hold not only the value of its quantity as a `KernelFunctionOperation`, but which would also keep track of the terms in their budget equation. So for example, for KE we'd have something like

```julia
struct KineticEnergy <: AbstractBudgetedQuantity
    quantity :: KFO
    shear_production
    pressure_flux
    buoyancy_flux
    stress
end
```

I guess the advantage of that over having a `KineticEnergyEquation` module is that we could do stuff like
```julia
interior(abq::AbstractBudgetedQuantity) = interior(abq.quantity)
```
plus other operations that are expected to work with KernelFunctionOperations.

Not sure if this would work, just spitballing here.

test/test_utils.jl
```diff
@@ -0,0 +1,67 @@
...
+arch = has_cuda_gpu() ? GPU() : CPU()
+
+# Grid size
+N = 6
+
+# Regular grid
+regular_grid = RectilinearGrid(arch, size=(N, N, N), extent=(1, 1, 1))
+
+# Stretched grid functions and setup
+S = .99 # Stretching factor. Positive number ∈ (0, 1]
```
glwagner, line 17:
```suggestion
S = 0.99 # Stretching factor. Positive number ∈ (0, 1]
```

docs/examples/kelvin_helmholtz.jl
```diff
@@ -163,6 +163,6 @@ end
 # ![](kelvin_helmholtz.mp4)
 #
 # Similarly to the kinetic energy dissipation rate (see the [Two-dimensional turbulence example](@ref two_d_turbulence_example)),
-# `TracerVarianceDissipationRate` and `TracerVarianceDiffusiveTerm` are implemented
+# `TracerVarianceDissipationRate` and `TracerVarianceDiffusion` are implemented
 # with a energy-conserving formulation, which means that (for `NoFlux` boundary conditions) their
```
jbisits, line 167: with an energy-conserving formulation, which means that for `NoFlux` boundary conditions their

jbisits (approved):

jbisits:
Big effort @tomchor, nice job! The reformatting of the modules into `...Equation` is great. I wonder if in a future PR the formula for, ```...Equation``` could be added as a docstring for the module as I think this would then be rendered in the documentation. This would provide quite a quick way for people to see what diagnostics you have written. Something like

```julia
"""
Module that contains the terms for the kinetic energy equation:

```math
\\frac{\partial}{\partial}E_{k} = -\\frac{u_{j} E_{k}}{\partial x_{j}} - \\frac{u_{i}p_{i}}{x_{i} + u_{i}b_{i} - u_{i}\\frac{\tau_{ij}}{x_{j}}
 
Short description of the terms and what functions they are called with.
"""
module TurbulentKineticEnergyEquation
```

test/test_kinetic_energy_equation.jl
```diff
@@ -0,0 +1,185 @@
...
+
+    u⃗∇pNHS = KineticEnergyEquation.PressureRedistribution(model, pressure=model.pressures.pNHS)
+    @test u⃗∇pNHS isa KineticEnergyEquation.KineticEnergyPressureRedistribution
+    @test Field(u⃗∇pNHS) isa Field
+
+    # Test calculation with a hydrostatic pressure separation
+    model2 = NonhydrostaticModel(grid=model.grid, hydrostatic_pressure_anomaly=CenterField(model.grid))
+    u⃗∇p_from_model2 = KineticEnergyEquation.PressureRedistribution(model2)
+    @test u⃗∇p_from_model2 isa KineticEnergyEquation.KineticEnergyPressureRedistribution
+    @test Field(u⃗∇p_from_model2) isa Field
```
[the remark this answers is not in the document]
tomchor: I think we have dug into this before. iirc what defines whether or not things are correctly printed is the font family that your browser uses, not github. For example, these appear fine for me on both firefox and chrome:

image

I'm using noto for firefox and Juliamono (for fixed-width) on chrome for the record. Depending on the font I choose, I also see squares.

src/FlowDiagnostics.jl
```diff
@@ -63,6 +63,8 @@ end
     return dbdz / duₕdz^2
 end
 
+const RichardsonNumber = KernelFunctionOperation{<:Any, <:Any, <:Any, <:Any, <:Any, <:typeof(richardson_number_ccf)}
```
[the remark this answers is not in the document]
tomchor: Nice! Done

docs/examples/kelvin_helmholtz.jl
tomchor: Sorry @jbisits not sure I follow your comment

PR #226, opened 2026-04-18
Add Filters submodule with BoxFilter

tomchor:
This PR implements a `Filters` module with one filter for now: `BoxFilter`. After this is merged I intend to implement a `GaussianFilter` as well (and perhaps more in the future as needed).

## Summary
- New `Filters` submodule exposing `BoxFilter(ψ; dims, width)`, a `KernelFunctionOperation`-based box-average filter over any subset of directions (1D, 2D, or 3D)
- `dims` is a tuple of integers drawn from `(1, 2, 3)`; `width` is the (mandatory) half-width of the symmetric running-mean stencil
- Output location matches input location; accepts any input supporting the `ψ[i, j, k]` indexing contract (`Field`, `KernelFunctionOperation`, other `AbstractOperation`s)
- Per-direction boundary handling for `Bounded` topologies via a `boundary` keyword (default: `:shrink`):
  - `:shrink` — near-wall stencils shrink, giving an honest local average
  - `:edge` — out-of-bounds offsets clamp to the nearest boundary cell
  - `(left=a, right=b)` — constant padding on each side
- `Periodic` directions always wrap via `mod1` regardless of the `boundary` spec
- No halo-size constraint: all policies resolve indices before reading, so `width` may exceed the halo
- Follows the same `#+++ / #---` section convention and `CustomKFO` alias idiom as the rest of the codebase

## Test plan
- [x] Constructor returns a `KernelFunctionOperation` / `BoxFilter`
- [x] Linear field unchanged on interior cells across all 1D / 2D / 3D `dims` combinations
- [x] Constant field unchanged
- [x] Periodic: numerical equivalence against an explicit mod1-wrapped stencil (1D, 2D, 3D, widths 1 and 2)
- [x] Periodic: hand-computed values on a 10-cell 1D grid (widths 1 and 2)
- [x] Bounded `:shrink`: cell-by-cell match against an explicit shrink reference (1D–3D, width 2)
- [x] Bounded `:edge`: cell-by-cell match against an explicit clamp reference (1D and 3D, width 2)
- [x] Bounded constant-pad: cell-by-cell match against an explicit reference (1D per-direction, widths 2 and 3)
- [x] Hand-computed values on a 10-cell 1D bounded grid for `:shrink`, `:edge`, and constant-pad at widths 2 and 3
- [x] Per-dim boundary tuple correctly mixes policies across dims
- [x] Periodic dims silently ignore the user-supplied `boundary` spec
- [x] Small-halo grids accepted for any width and topology
- [x] Output location matches input location for ccc / fcc / cfc / ccf inputs
- [x] Accepts a `BinaryOperation` and a nested `KernelFunctionOperation` as input
- [x] Rejects invalid `dims`, `width`, and `boundary` specs at construction time
- [x] Test suite passes: `TEST_GROUP=filters julia --project=. -e 'using Pkg; Pkg.test()'` → 68/68 pass

### DIFF
--- a/src/Filters.jl
+++ b/src/Filters.jl
@@ -69,9 +69,18 @@ ConstantBoundary(left, right) = ConstantBoundary{promote_type(typeof(left), type
 @inline _fetch2(c::ConstantBoundary, ψ, i, j, k, N) = (ifelse(j < 1, c.left, ifelse(j > N, c.right, @inbounds ψ[i, clamp(j, 1, N), k])), 1)
 @inline _fetch3(c::ConstantBoundary, ψ, i, j, k, N) = (ifelse(k < 1, c.left, ifelse(k > N, c.right, @inbounds ψ[i, j, clamp(k, 1, N)])), 1)
 
-@inline _fetch1(::ShrinkBoundary, ψ, i, j, k, N) = (1 <= i <= N) ? (@inbounds ψ[i, j, k], 1) : (zero(eltype(ψ)), 0)
-@inline _fetch2(::ShrinkBoundary, ψ, i, j, k, N) = (1 <= j <= N) ? (@inbounds ψ[i, j, k], 1) : (zero(eltype(ψ)), 0)
-@inline _fetch3(::ShrinkBoundary, ψ, i, j, k, N) = (1 <= k <= N) ? (@inbounds ψ[i, j, k], 1) : (zero(eltype(ψ)), 0)
+@inline function _fetch1(::ShrinkBoundary, ψ, i, j, k, N)
+    in_bounds = (1 <= i) & (i <= N)
+    (ifelse(in_bounds, @inbounds(ψ[clamp(i, 1, N), j, k]), zero(eltype(ψ))), Int(in_bounds))
+end
+@inline function _fetch2(::ShrinkBoundary, ψ, i, j, k, N)
+    in_bounds = (1 <= j) & (j <= N)
+    (ifelse(in_bounds, @inbounds(ψ[i, clamp(j, 1, N), k]), zero(eltype(ψ))), Int(in_bounds))
+end
+@inline function _fetch3(::ShrinkBoundary, ψ, i, j, k, N)
+    in_bounds = (1 <= k) & (k <= N)
+    (ifelse(in_bounds, @inbounds(ψ[i, j, clamp(k, 1, N)]), zero(eltype(ψ))), Int(in_bounds))
+end
 
 @inline _call1(::PeriodicBoundary, f, i, j, k, N, grid, fargs...) = (f(mod1(i, N), j, k, grid, fargs...), 1)
 @inline _call2(::PeriodicBoundary, f, i, j, k, N, grid, fargs...) = (f(i, mod1(j, N), k, grid, fargs...), 1)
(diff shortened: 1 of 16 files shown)


tomchor:
An example of the result of BoxFilter in a Kelvin Helmholtz simulation from the example. Original buoyancy field:

Screenshot_20260420_080947

Filtered field (with length scale ~ 2):

Screenshot_20260420_081044

In the example above the filter is periodic in the horizontal dir (matches the boundary condition) and has a shrinking stencil as it approches the boundary. Other options are `:edge` boundary policies (where the last value of the domain is used to  pad the results) and padding with specified values for the left and right boundaries.

src/Filters.jl
```diff
@@ -0,0 +1,364 @@
...
+@inline _fetch2(::PeriodicBoundary, ψ, i, j, k, N) = (ψ[i, mod1(j, N), k], 1)
+@inline _fetch3(::PeriodicBoundary, ψ, i, j, k, N) = (ψ[i, j, mod1(k, N)], 1)
+
+@inline _fetch1(::EdgeBoundary, ψ, i, j, k, N) = (ψ[clamp(i, 1, N), j, k], 1)
+@inline _fetch2(::EdgeBoundary, ψ, i, j, k, N) = (ψ[i, clamp(j, 1, N), k], 1)
+@inline _fetch3(::EdgeBoundary, ψ, i, j, k, N) = (ψ[i, j, clamp(k, 1, N)], 1)
+
+@inline function _fetch1(c::ConstantBoundary, ψ, i, j, k, N)
+    i < 1 && return (c.left,  1)
+    i > N && return (c.right, 1)
```
glwagner: its best to avoid short circuiting logic in kernel functions, which I suspect this is due to `i, j, k` arguments

src/Filters.jl
```diff
@@ -0,0 +1,364 @@
...
+#
+# `_fetchD(policy, ψ, i, j, k, N)` returns `(value, count)` when `ψ` is an
+# indexable field; `_callD(policy, f, i, j, k, N, grid, fargs...)` is the
+# same thing when the inner value comes from another kernel function. `count`
+# is 1 in every case except `ShrinkBoundary` out-of-bounds, where it is 0 so
+# the offset is excluded from the running mean.
+
+@inline _fetch1(::PeriodicBoundary, ψ, i, j, k, N) = (ψ[mod1(i, N), j, k], 1)
+@inline _fetch2(::PeriodicBoundary, ψ, i, j, k, N) = (ψ[i, mod1(j, N), k], 1)
+@inline _fetch3(::PeriodicBoundary, ψ, i, j, k, N) = (ψ[i, j, mod1(k, N)], 1)
```
glwagner: do these need either `@inbounds` or `@propagate_inbounds`?

src/Filters.jl
```diff
@@ -0,0 +1,364 @@
...
+        val, cnt = _fetch1(policy, ψ, i + di, j, k, Nx)
+        s += val; n += cnt
+    end
+    return s / n
+end
+
+@inline function (::BoxFilterKernel{2})(i, j, k, grid, width, policy, ψ)
+    Ny = size(grid, 2)
+    s = zero(grid); n = 0
+    @inbounds for dj in -width:width
```
glwagner: you may need to pass in `Val(width)` if you want this to be unrolled

glwagner (commented):
this looks interesting! I think there are some performance concerns:

* missing `@inbounds` or `@propagate_inbounds`
* short circuiting logic that shows up in kernel
* need to unroll loops over the filter width; otherwise explicit loops will appear in compiled code, evaluated at every grid point

src/Filters.jl
tomchor: Thanks, I replaced all `&&` with either `&` or `ifelse` statements. Let me know if you had a different solution in mind

src/Filters.jl
tomchor: Added some `@inbounds`!

src/Filters.jl
```diff
@@ -65,65 +65,29 @@ ConstantBoundary(left, right) = ConstantBoundary{promote_type(typeof(left), type
...
-    return (f(i, j, k, grid, fargs...), 1)
-end
-@inline function _call3(c::ConstantBoundary, f, i, j, k, N, grid, fargs...)
-    k < 1 && return (c.left,  1)
-    k > N && return (c.right, 1)
-    return (f(i, j, k, grid, fargs...), 1)
-end
+@inline _fetch1(::ShrinkBoundary, ψ, i, j, k, N) = (1 <= i <= N) ? (ψ[i, j, k], 1) : (zero(eltype(ψ)), 0)
+@inline _fetch2(::ShrinkBoundary, ψ, i, j, k, N) = (1 <= j <= N) ? (ψ[i, j, k], 1) : (zero(eltype(ψ)), 0)
+@inline _fetch3(::ShrinkBoundary, ψ, i, j, k, N) = (1 <= k <= N) ? (ψ[i, j, k], 1) : (zero(eltype(ψ)), 0)
```
glwagner: the ternary operator is also short circuiting

src/Filters.jl
```diff
@@ -65,65 +65,29 @@ ConstantBoundary(left, right) = ConstantBoundary{promote_type(typeof(left), type
...
-    return (ψ[i, j, k], 1)
-end
-@inline function _fetch3(c::ConstantBoundary, ψ, i, j, k, N)
-    k < 1 && return (c.left,  1)
-    k > N && return (c.right, 1)
-    return (ψ[i, j, k], 1)
-end
+@inline _fetch1(c::ConstantBoundary, ψ, i, j, k, N) = (ifelse(i < 1, c.left, ifelse(i > N, c.right, ψ[clamp(i, 1, N), j, k])), 1)
+@inline _fetch2(c::ConstantBoundary, ψ, i, j, k, N) = (ifelse(j < 1, c.left, ifelse(j > N, c.right, ψ[i, clamp(j, 1, N), k])), 1)
+@inline _fetch3(c::ConstantBoundary, ψ, i, j, k, N) = (ifelse(k < 1, c.left, ifelse(k > N, c.right, ψ[i, j, clamp(k, 1, N)])), 1)
```
glwagner: this still needs either `@inbounds` or `@propagate_inbounds`

src/Filters.jl
tomchor: This is fixed. Somehow you commented on an out-of-date version of the code

src/Filters.jl
tomchor: Same here. This was fixed. Not sure how you were able to comment on old code. Maybe a weird github bug...?

src/Filters.jl
tomchor: Done!

src/Filters.jl
glwagner: hmm ok

tomchor:
Okay I think this is ready for re-review. @glwagner I believe I addressed all your points regarding GPU performance, and I also added docs (thanks, Claude!)

glwagner (approved):
looks good. not sure if you enforce this, but you are missing `return` statements. I also do not think it is necessary to explicitly return a tuple, eg if you write `return a, b` it is the same as `return (a, b)`.

also in Oceananigans we reserve `_` prefix for kernels; here it is used more generally as Claude likes to do to mean "private" function. Matter of taste.

tomchor:
> looks good. not sure if you enforce this, but you are missing `return` statements. I also do not think it is necessary to explicitly return a tuple, eg if you write `return a, b` it is the same as `return (a, b)`.
> 
> also in Oceananigans we reserve `_` prefix for kernels; here it is used more generally as Claude likes to do to mean "private" function. Matter of taste.

Thanks! I haven't enforced any of these things (and historically I have also used a underscore prefix to mean "private" function), but we can start enforcing those if you think it helps getting more contributions from Oceananigans devs. (So far it's been mainly just in this repo...)

I'll enforce these on this PR and then we can discuss enforcing these more generally from now on.

PR #229, opened 2026-04-27
Add GaussianFilter to Filters module

tomchor:
## Summary

Adds `GaussianFilter` alongside `BoxFilter` and reworks the user-facing filter API.

### New filter
- `GaussianFilter` — Gaussian-weighted local average using `exp(-Δ²/(2σ²))` with runtime normalization for consistent boundary handling
- `σ` (mandatory keyword) is in **physical units** (same as grid spacing); per-direction stencil sizes are inferred from `σ` and `minimum_xspacing/yspacing/zspacing` so the stencil extends ~2σ on each side
- Per-direction widths and weights, so the kernel adapts correctly to anisotropic uniform grids
- Boundary policies (`:shrink`, `:edge`, `(left, right)`) are shared with `BoxFilter`
- **Uniform-spacing check**: the per-direction weights are precomputed assuming a single Δ per direction, so `GaussianFilter` rejects non-uniform (stretched) filtered directions at construction time with an `ArgumentError` pointing the user toward `BoxFilter`. Other (non-filtered) directions on the same grid may still be stretched.

### API change: `width`/`n_points` → `N`
Both filters now take a single keyword `N` — the **total number of grid points used by the filter stencil** along each filtered direction. `N` must be an **odd integer ≥ 3** so the stencil is symmetric around the current cell. Internal kernels still use the half-width `(N - 1) ÷ 2`. For `Periodic` directions the stencil must span at most one period (`N ≤ 2*Nd_grid + 1`), enforced at construction time.

### Code organization
- Split `src/Filters.jl` into a `src/Filters/` directory with snake_case files: `Filters.jl` (shared infrastructure), `box_filter.jl`, `gaussian_filter.jl`
- Shared helpers: `resolve_filter_policies`, `build_filter_kfo`, `validate_N`, `validate_periodic_widths`
- GaussianFilter-specific helpers (`direction_min_spacing`, `direction_spacings`, `validate_uniform_spacing`, `infer_width`) live in `gaussian_filter.jl`
- Stencil readers renamed `_fetch1/2/3` → `x/y/z_stencil_fetch` (and the `_call` variants), with brief documentation of the `(value, count)` return convention

### GPU performance
- Loop bounds passed via `Val(width)` so the symmetric stencil loop unrolls at compile time
- `@inbounds` on every array access in the stencil readers
- ShrinkBoundary uses branchless `ifelse` + `clamp` instead of ternary `?:` (no warp divergence)
- `mod1` replaced with a branchless conditional wrap (no GPU integer-division stall)

### Tests
- Both filters share a parametrized test suite, grouped into a few logical testsets (smoke, identity properties, numerical correctness, composability, argument validation)
- `compute_filter` test helper takes the half-width and converts to `N = 2*width + 1` in one place
- Dirac-delta impulse-response test for `GaussianFilter` — confirms the filter's response to a δ is the normalized Gaussian kernel itself, across multiple `(N, σ)` combinations
- Inference tests: uniform grid (`test_gaussian_N_inferred`) and anisotropic uniform grid `Δx=0.125, Δz=0.25` (`test_gaussian_N_inferred_anisotropic`) — verify auto-inferred per-direction `N` equals the explicit value
- Tuple-dim ordering test (`test_gaussian_N_tuple_order`): confirms `dims=(3, 1), N=(5, 7)` matches `dims=(1, 3), N=(7, 5)` and rejects mismatched length, even values, below-min values, and non-integer entries
- Uniform-spacing test (`test_gaussian_uniform_spacing_required`): on a stretched z-grid, filtering z must error, filtering x/y still works, and `BoxFilter` is unaffected
- Periodic constraint test (`test_periodic_N_too_large`): verifies `N > 2*Nd_grid + 1` errors on periodic dims and succeeds on bounded dims
- Hand-computed 1D tests now use `copyto!` (GPU-safe) instead of broadcast `.=` for the initial fill
- Test count: 68 → 166

### Docs
- New `docs/src/filters.md` covering both filters, the boundary handling table, default-vs-override behavior for `N`, per-dim usage, and a `!!! warning` admonition about the GaussianFilter uniform-spacing requirement
- All examples on the page are `jldoctest` blocks with verified outputs (`size`, `isa Field`)
- Filter docstrings updated with bold emphasis on the meaning of `N` and an aside clarifying it is the *filter stencil size, not the grid size*

## Test plan
- [x] All 166 filter tests pass on CPU
- [x] Doctests pass (page-level + docstring-level)
- [x] Verify CI passes on all test groups


src/Filters/Filters.jl
```diff
@@ -0,0 +1,217 @@
...
+# excluded from the running mean.
+#
+# `x/y/z_stencil_fetch` read from an indexable field `ψ`.
+# `x/y/z_stencil_call`  evaluate a kernel function `f` at the adjusted index.
+
+@inline wrap_periodic_index(i, N) = i + ifelse(i < 1, N, 0) - ifelse(i > N, N, 0)
+
+@inline x_stencil_fetch(::PeriodicBoundary, ψ, i, j, k, N) = (@inbounds ψ[wrap_periodic_index(i, N), j, k], 1)
+@inline y_stencil_fetch(::PeriodicBoundary, ψ, i, j, k, N) = (@inbounds ψ[i, wrap_periodic_index(j, N), k], 1)
+@inline z_stencil_fetch(::PeriodicBoundary, ψ, i, j, k, N) = (@inbounds ψ[i, j, wrap_periodic_index(k, N)], 1)
```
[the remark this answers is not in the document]
tomchor: @copilot apply changes based on [this feedback](https://github.com/tomchor/Oceanostics.jl/pull/229#discussion_r3177408301) Validate such that n_points > 2N+1

src/Filters/gaussian_filter.jl
```diff
@@ -0,0 +1,172 @@
...
+            throw(ArgumentError("`n_points` tuple must have one entry per dim in `dims`; got length $(length(n_points)) for dims=$dims"))
+        foreach(validate_n_points, n_points)
+        return ntuple(i -> begin
+            user_idx = findfirst(==(sorted_dims[i]), dims)
+            (n_points[user_idx] - 1) ÷ 2
+        end, length(sorted_dims))
+    else
+        validate_n_points(n_points)
+        return ntuple(_ -> (n_points - 1) ÷ 2, length(sorted_dims))
+    end
```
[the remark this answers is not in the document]
tomchor: @copilot apply changes based on [this feedback](https://github.com/tomchor/Oceanostics.jl/pull/229#discussion_r3177408327)

src/Filters/Filters.jl
```diff
@@ -0,0 +1,226 @@
...
+    in_bounds = (1 <= i) & (i <= N)
+    return ifelse(in_bounds, f(clamp(i, 1, N), j, k, grid, fargs...), zero(grid)), Int(in_bounds)
+end
+@inline function y_stencil_call(::ShrinkBoundary, f, i, j, k, N, grid, fargs...)
+    in_bounds = (1 <= j) & (j <= N)
+    return ifelse(in_bounds, f(i, clamp(j, 1, N), k, grid, fargs...), zero(grid)), Int(in_bounds)
+end
+@inline function z_stencil_call(::ShrinkBoundary, f, i, j, k, N, grid, fargs...)
+    in_bounds = (1 <= k) & (k <= N)
+    return ifelse(in_bounds, f(i, j, clamp(k, 1, N), grid, fargs...), zero(grid)), Int(in_bounds)
```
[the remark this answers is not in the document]
tomchor: This is most performant option for GPU kernels

docs/src/filters.md
```diff
@@ -9,26 +9,29 @@ other operations, etc.) and run on both CPU and GPU.
 ## Box filter
 
 The [`BoxFilter`](@ref) computes a local running-mean (box average) of a field
-over one or more grid directions. The stencil width is controlled by the `width`
-keyword: each filtered direction uses a `(2*width + 1)`-point symmetric average
-centred on the current cell.
+over one or more grid directions. The stencil size is controlled by the
+`n_points` keyword: each filtered direction uses an `n_points`-point symmetric
```
glwagner: `Npoints`? follow Oceananigans style?

src/Filters/gaussian_filter.jl
glwagner, line 1: this file should define a module, or be renamed to `gaussian_filter.jl`. same for others

glwagner: this is Julia-recommended style

tomchor: Ah, thanks for catching that. Made the change

docs/src/filters.md
tomchor: I didn't realize Oceananigas had a style for this sort of stuff. `Npoints` isn't really PascalCase or cameCase or anything though. And usually arguments follow snake_case, so I'm confused. Can you please clarify which case `Npoints` falls into?

src/Filters/Filters.jl
glwagner:
```julia
@inline wrap_periodic_index(i, N) = i + N * (i < 1) - N * (i > N)
```

no big deal, but I think easier to read

tomchor: I agree. Just made that change. I'll merge after tests pass

docs/src/filters.md
tomchor: I'll wait for you to respond to this today, otherwise I'll merge later in the day. I feel like it's good to keep the standard for function arguments that currently are always snake_case

glwagner: It doesn't call into a case. We use it for all "number of" things. `Nx`, `Ny`, `Nz`, `Nt`.

glwagner: I recommend keeping this style

tomchor: Yeah but `N` I think is already taken to mean `number of points`. Like, `Nx` to me "parses" as  the `number of points in x`. Conversely `Npoints` feels redundant to me. For whatever reason `n_points` doesn't even though it's the same letter lol

Maybe just `N`? The docstring already makes it clear what it means, and that way it's inline with Oceananigans.

tomchor: Went with `N` since it follows Ocenanigans and it's simpler

PR #244, opened 2026-06-04
Add StrainRateTensor diagnostic

tomchor:
Adds `StrainRateTensor`, returning the independent components of the strain rate tensor `Sᵢⱼ = ½(∂ⱼuᵢ + ∂ᵢuⱼ)` as a `NamedTuple` of KFOs at their natural staggered locations. A `dims` kwarg (default `(1, 2, 3)`) selects sub-dimensional tensors; e.g. `dims=(1, 3)` returns the 2D `x`–`z` tensor (`S₁₁`, `S₃₃`, `S₁₃`). Complements `StrainRateTensorModulus`. Tested and documented.

### DIFF
--- a/docs/src/flow_diagnostics.md
+++ b/docs/src/flow_diagnostics.md
@@ -153,6 +153,24 @@ Oceanostics.FlowDiagnostics.StrainRateTensor
 Oceanostics.FlowDiagnostics.StrainRateTensorModulus
 ```
 
+### Principal strain rates
+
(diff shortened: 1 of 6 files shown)


tomchor:
Here's a comparison with an offline calculation via xarray (on a Kelvin Helmholtz simulation):

khi_Nz128_Ri0 10_strain_tensor_comparison_l7_t30 0

Note that I'm not interpolating correctly in the xarray code (especially not at the boundaries) so this difference is expected. The difference between both does decrease with a higher resolution (also as expected).

PR #248, opened 2026-06-07
(0.17.1) Add CoarseGrainedKineticEnergyEquation module (with cross-scale KE flux)

tomchor:
## Summary

Adds a new `CoarseGrainedKineticEnergyEquation` module implementing the full **3D** cross-scale (scale-to-scale) kinetic-energy flux that appears in the filtered/coarse-grained KE budget — a generalization of the 2D (x–z) metric used in offline postprocessing.

Two public diagnostics:

- **`subfilter_stress_tensor(model, filter; dims, collocate_diagonals)`** — the subfilter-scale stress tensor `τⁱʲ = filter(uⁱuʲ) − ūⁱūʲ`, returned as a `NamedTuple` of components that mirror `StressTensor` (same locations; `dims` and `collocate_diagonals` forwarded).
- **`KineticEnergyCrossScaleFlux(model, filter; dims)`** — the cross-scale flux `Πₖ = −τⁱʲ S̄ⁱʲ` (Aluie et al., 2018, [doi:10.1175/JPO-D-17-0100.1](https://doi.org/10.1175/JPO-D-17-0100.1)), contracted at cell centers with off-diagonals counted twice. `Πₖ > 0` is forward (downscale) transfer; the result is per unit mass.

```julia
ℓ = 0.2  # filter FWHM
filter = GaussianFilter(; dims=(1,2,3), σ=ℓ/(2√(2log(2))), boundary=:edge)   # reusable filter object
Πₖ = KineticEnergyCrossScaleFlux(model, filter)
# or the convenience form, which builds the Gaussian filter from σ:
Πₖ = KineticEnergyCrossScaleFlux(model; σ=ℓ/(2√(2log(2))), boundary=:edge)
```

## Design

- **Reuses existing building blocks**: `S̄ⁱʲ` is just `StrainRateTensor` of the filtered velocity; `τⁱʲ` is `StressTensor` of the full velocity, filtered, minus `StressTensor` of the filtered velocity. The only genuinely new code is the subfilter-stress assembly and the co-located contraction.
- **Filter as any callable**: pass a reusable `GaussianFilter(; ...)` (or any field→field callable). This decouples the filter’s own directions/boundary/scale from the contraction `dims` (so you can filter horizontally yet contract the full tensor). A convenience method builds the Gaussian filter from `σ` for you.
- **Location in the tree**: included after `FlowDiagnostics` (tensors) and `SpatialFilters` (GaussianFilter), the two modules it depends on.
- **Lazy but fast**: filtered velocities and momentum fluxes are materialized as `Field`s internally (the separable filter’s staged path only fires when wrapped directly in a `Field`), while the returned `Πₖ` stays a lazy operation that recomputes correctly in an `OutputWriter`.

## Testing

New `coarse_grained_ke_diagnostics` test group (+ CI job):

- structural: component keys, locations match `StressTensor`, `dims` subsetting, `collocate_diagonals` moves diagonals to `ccc`, invalid `dims` rejected;
- **correctness**: `Πₖ` matches an independent by-hand contraction (guards the components, the ×2 on off-diagonals, and the sign);
- **uniform flow** ⇒ `τⁱʲ = 0` and `Πₖ = 0` to machine precision;
- a reusable `GaussianFilter` object is accepted directly as `filter`, matching the equivalent closure;
- the diagnostic recomputes correctly as the flow evolves (re-`compute_at!` reflects updated velocities);
- the `GaussianFilter` convenience method reproduces the explicit filter call.

Also adds a docs page and `CLAUDE.md` updates.


src/CoarseGrainedKineticEnergyEquation.jl
```diff
@@ -0,0 +1,189 @@
...
+full tensor.
+
+A convenience method `CrossScaleKineticEnergyFlux(model; σ, dims, boundary, N)` builds the Gaussian
+`filter` for you from a standard deviation `σ` (with `σ = ℓ / (2√(2 ln 2))` for a FWHM `ℓ`).
+"""
+function CrossScaleKineticEnergyFlux(model, filter; dims = (1, 2, 3))
+    FlowDiagnostics.validate_dims(dims)
+    grid = model.grid
+    u, v, w = model.velocities
+    ū, v̄, w̄ = _filtered_velocities(filter, dims, u, v, w)
```
glwagner: Just a note  that we use the term "filter" in Breeze for time filtering. So without qualification, "filter" can be ambiguous (spatial, temporal, or both)

src/CoarseGrainedKineticEnergyEquation.jl
```diff
@@ -0,0 +1,189 @@
...
+
+    # Resolved-scale strain S̄ⁱʲ from the filtered velocities, and the subfilter stress τⁱʲ. The
+    # contraction below interpolates every component to cell centers, so the components can stay at
+    # their natural staggered locations here.
+    S̄ = StrainRateTensor(grid, ū, v̄, w̄; dims)
+    τ = _subfilter_stress_tensor(filter, grid, u, v, w, ū, v̄, w̄; dims)
```
glwagner: should you remove the `_` prefix?

src/CoarseGrainedKineticEnergyEquation.jl
```diff
@@ -0,0 +1,189 @@
...
+
+julia> model = NonhydrostaticModel(grid);
+
+julia> ℓ = 0.2; # filter scale (FWHM)
+
+julia> filt = ψ -> GaussianFilter(ψ; dims=(1, 2, 3), σ=ℓ / (2√(2log(2))));
+
+julia> Πₖ = CrossScaleKineticEnergyFlux(model, filt);
+
+julia> Field(Πₖ) isa Field
```
glwagner: Here, i think you should just show the output of `CrossScaleKineticEnergyFlux`. I also think it could make sense to use a code block instead of REPL form. Finally, why can't `GaussianFilter` be passed directly to `CrossScaleKineticEnergyFlux`?

src/CoarseGrainedKineticEnergyEquation.jl
```diff
@@ -0,0 +1,189 @@
...
+#---
+
+#+++ Cross-scale kinetic-energy flux
+# Πₖ = -τⁱʲ S̄ⁱʲ contracted at cell centers. τ and S̄ share keys/ordering (both built with the same
+# `dims`), so we pair them component-by-component and weight the off-diagonals by 2 (tensor symmetry).
+# Each component is interpolated to (Center, Center, Center) before multiplying, matching the offline
+# postprocessing convention.
+to_center(ψ) = @at (Center, Center, Center) ψ
+
+const _CONTRACTION = ((:τ₁₁, :S₁₁, 1), (:τ₂₂, :S₂₂, 1), (:τ₃₃, :S₃₃, 1),
```
glwagner, line 114: what does the `_` mean here?

docs/src/coarse_grained_kinetic_energy_equation.md
```diff
@@ -0,0 +1,61 @@
...
+```julia
+using Oceananigans, Oceanostics
+
+grid = RectilinearGrid(size=(16, 16, 16), extent=(1, 1, 1), topology=(Periodic, Periodic, Bounded))
+model = NonhydrostaticModel(grid)
+set!(model, u=(x, y, z) -> randn(), v=(x, y, z) -> randn(), w=(x, y, z) -> randn())
+
+# A Gaussian filter of full width at half maximum ℓ in all three directions
+ℓ = 0.2
+filter = ψ -> GaussianFilter(ψ; dims=(1, 2, 3), σ=ℓ / (2√(2log(2))), boundary=:edge)
```
glwagner:
```suggestion
filter = GaussianFilter(ψ; dims=(1, 2, 3), σ=ℓ / (2√(2log(2))), boundary=:edge)
```

is this syntax possible? it is a bit shorter and perhaps easier to remember and read

docs/src/coarse_grained_kinetic_energy_equation.md
```diff
@@ -0,0 +1,61 @@
...
+
+Πₖ = CrossScaleKineticEnergyFlux(model, filter)   # the cross-scale KE flux, at (Center, Center, Center)
+τ  = SubfilterStressTensor(model, filter)         # the subfilter stress tensor components
+
+# the convenience methods build the Gaussian filter from σ for you:
+Πₖ = CrossScaleKineticEnergyFlux(model; σ=ℓ / (2√(2log(2))), boundary=:edge)
```
glwagner: the syntax `boundary=:edge` isn't clear to me

glwagner (commented):
mostly syntax comments

src/CoarseGrainedKineticEnergyEquation.jl
tomchor: Hmm, maybe we could rename the `Filters` module to `SpatialFilters`? That might be better done in another PR though since it's unrelated to this one

src/CoarseGrainedKineticEnergyEquation.jl
tomchor: I actually like leading `_` for functions that are supposed to be only internal. But I can remove it if you feel strongly about it

src/CoarseGrainedKineticEnergyEquation.jl
tomchor: It just signifies that this is for internal use

docs/src/coarse_grained_kinetic_energy_equation.md
tomchor: It's explained in the docstring for `BoxFilter`, but you're right that maybe it isn't very intuitive. Here's what it means:

```
  Stencil offsets that leave the interior 1:Nd_grid of a direction (where Nd_grid is the number of cells along that direction) are handled per-direction. For Periodic directions offsets
  are always wrapped periodically, independent of the boundary keyword. For Bounded directions the boundary keyword picks the policy (default: :shrink):

    •  :shrink — drop out-of-bounds offsets from both the sum and the count, so the filter is an honest local average whose effective stencil shrinks near a wall. This is the default
       for Bounded directions.
    •  :edge — replicate the boundary-cell value (reads ψ[1] or ψ[Nd_grid] for offsets past either end).
    •  (left=a, right=b) — pad with constant a on the low-index side and b on the high-index side (a and b are promoted to a common type).
```

tomchor: Maybe for now I'll just remove `boundary=:edge` from the example and open an issue to discuss a possible better name for this option.

docs/src/coarse_grained_kinetic_energy_equation.md
tomchor: No but I changed it to `filter(ψ) = GaussianFilter(ψ; dims=(1, 2, 3), σ=ℓ / (2√(2log(2))), boundary=(left=0, right=0))`, which is more readable. Also made this a script-style jldoctest

docs/src/coarse_grained_kinetic_energy_equation.md
tomchor: Replaced it with `boundary=(left=0, right=0)` which is clearer. Also opened https://github.com/tomchor/Oceanostics.jl/issues/251 to discuss this notation

src/CoarseGrainedKineticEnergyEquation.jl
tomchor: Done and done. Lastly, at the moment `GaussianFilter` can't be passed directly because it takes two arguments: a field to be filtered and a scale. I guess we could pass a `Filter` directly, but we'd need to also enforce passing a scale, which I think kills the simplicity of this type of interface.

tomchor: But that's something we can change. Especially since, now that I think about it, I think it'd lead to a better user experience. Maybe something like `gf = GaussianFilter(; dims=(1,2,3), σ=0.1)` could return an operation.

docs/src/coarse_grained_kinetic_energy_equation.md
glwagner: what does `boundary` mean here? (what does the kwarg specify?)

tomchor: It's the behavior of the filter at the boundaries. The keyword comes from the filter, so if we're gonna change it, we should do it there.

src/CoarseGrainedKineticEnergyEquation.jl
tomchor: Changed

src/CoarseGrainedKineticEnergyEquation.jl
tomchor: Renamed to SpatialFilters

tomchor:
After the latest changes, `subfilter_stress_tensor` is a single function with three methods (dispatched by signature):

- `subfilter_stress_tensor(model, filter; dims, collocate_diagonals)` — model-based
- `subfilter_stress_tensor(model; σ, dims, boundary, N, collocate_diagonals)` — convenience (builds the Gaussian filter from \`σ\`)
- `subfilter_stress_tensor(filter, grid, u, v, w, ū, v̄, w̄; dims, collocate_diagonals)` — low-level, takes already-filtered velocities (shared with \`KineticEnergyCrossScaleFlux\` so the velocities are filtered once)

PR #250, opened 2026-06-07
Add per-diagnostic show methods on top of KernelFunctionOperation

tomchor:
## What

Each diagnostic in Oceanostics is a `KernelFunctionOperation` (KFO) under the hood, so they all displayed as the generic `KernelFunctionOperation at (…)`. This gives every diagnostic its own `show` that **expands on** the underlying KFO's:

```
RossbyNumber KernelFunctionOperation at (Face, Face, Face)
├── grid: 4×4×4 RectilinearGrid{…} on CPU with 3×3×3 halo
├── kernel_function: rossby_number_fff (generic function with 1 method)
└── arguments: ("Field", "Field", "Field", "NamedTuple")
└── computes: Rossby number  Ro = ωᶻ/f
```

On a color-capable terminal the leading diagnostic name and the `computes:` line are tinted (configurable via `DESCRIPTION_CRAYON` / `set_description_color!`); plain/captured streams (pipes, files, doctests) print without escapes.

## How

A single `@diagnostic_show` macro (centralized in `src/Oceanostics.jl`) registers, per diagnostic:

1. an `operation_name` method renaming the header to `"<Name> KernelFunctionOperation"` — the `KernelFunctionOperation` suffix is kept so it stays clear the object is still a KFO. Because it goes through Oceananigans' `operation_name`, the rename also propagates to `summary(op)`, to operation trees, and to `Field` operand lines.
2. a `Base.show` method that delegates the grid/kernel/arguments tree to Oceananigans' KFO `show` via `invoke` and appends a `└── computes: …` description line.

**63 diagnostics** across all modules are registered.

## Notes / intentionally not covered

- Generic `*Forcing` aliases keep the default display (the kernel is a user-supplied forcing function — no fixed type to dispatch on).
- `StrainRateTensor` / `VorticityTensor` / `StressTensor` return `NamedTuple`s of raw KFOs, so their components stay generic.
- TKE's `IsotropicDissipationRate` reuses the KE kernel, so it displays as `KineticEnergyIsotropicDissipationRate`.

## Doctests

Changing `show` touches every `jldoctest` that printed the old header. All blocks in the docstrings (`src/*.jl`) and the `docs/src/*.md` pages were regenerated to match the new output (the `.md` blocks need a separate `makedocs(...; doctest=:fix)` pass). The `general_flow_diagnostics` test group also passes. Some incidental Oceananigans method-count refreshes (e.g. `10 methods → 4`) are cosmetic — normalized by the existing `doctestfilter`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,3 +76,4 @@ The `perf_invariants` test group guards against performance regressions without 
 - One-line code expressions are preferred when they fit within 130 columns; only break them across lines when they exceed that width
 - Prose text (docstrings, comments, `.md` files) should wrap at around 100 columns
 - When adding a new leaf progress messenger, wrap its formatted-number string (the result of `@sprintf` / `prettytime`) in `ColoredNumber(...)` so the value participates in the configurable `NUMBER_CRAYON` coloring; prefix and unit text stay as plain `String`
+- **Code folding markers**: collapsible code sections are delimited by `#+++ <title>` to open (note the space after `#+++`) and `#---` to close — always exactly three `+`/`-`, never `#++`/`#--`. Nested sections use the same `#+++`/`#---` markers (each `#---` closes the most recent `#+++`)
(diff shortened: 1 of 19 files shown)


tomchor:
For the record, currently the show method in this PR has some color and it's looking like:

image

PR #256, opened 2026-06-24
Match Kelvin-Helmholtz example to standard KH nondimensionalization

tomchor:
Updates the nondimensionalization in `docs/examples/kelvin_helmholtz.jl` to use the stratified-shear-layer scaling of [Kaminski & Smyth (2019)](https://doi.org/10.1017/jfm.2018.973), as used in [CoarseGrainedKHAPE](https://github.com/tomchor/CoarseGrainedKHAPE). The previous version was (imo) a bit confusing the length scale and  more importantly (I think) "squeezed" the instability by forcing it to happen in a domain that's shorter than the most unstable wavenumber. This PR changes it such that the domain length is approximately equal to 1 most-unstable-wavenumber.

Text below was written by Claude, so take it with a grain of salt.

## Summary of changes

| | Before | After (this PR) |
|---|---|---|
| Viscosity / diffusivity | hard-coded `ν = κ = 2×10⁻⁵` | explicit `Re`, `Pr`; `ν = U h / Re`, `κ = ν / Pr` |
| Shear `u` | `tanh(z) + noise` | `U·tanh(z/h)` (clean) |
| Stratification `b` | `h·Ri₀·tanh(z/h)`, `h = 1/4` (4× thinner than the shear) | `B₀·tanh(z/h)`, `B₀ = U² Ri₀ / h` (same length scale as the shear) |
| Richardson number | `Ri₀ = 0.1` (*maximum* gradient Ri) | `Ri₀ = 0.1` (*minimum* gradient Ri) |
| Domain | `10 × 10` | `Lx = 2π/k_max ≈ 14.1` (one most-unstable wavelength), `Lz = 10` |
| Perturbation | white noise on `u` (`2×10⁻²·randn()`) | structured on `w`: `0.05·abs(randn())·exp(−z²)·sin(k_max·x − π)`, `k_max = 0.4446/h` |
| `stop_time` | `100` | `120` |

## Nondimensionalization

Length is scaled by the shear-layer half-width `h` and velocity by `U`, so the flow is governed by the Richardson number `Ri₀`, the Reynolds number `Re = U h / ν`, and the Prandtl number `Pr = ν / κ`. The shear and the stratification share the same length scale `h`, and the buoyancy jump `B₀ = U² Ri₀ / h` makes `Ri₀` the minimum gradient Richardson number `N²/(∂u/∂z)²`, attained at the center of the shear layer where the flow is most unstable.

## Initial condition and domain

The shear and stratification are now clean `tanh` profiles, and the instability is seeded by perturbing the **vertical velocity** with the most unstable mode, `w = 0.05·abs(randn())·exp(−z²)·sin(k_max·x − π)`, where `k_max = 0.4446/h` is the most unstable Kelvin-Helmholtz wavenumber ([Michalke, 1964](https://doi.org/10.1017/S0022112064000908)). The horizontal domain is set to one wavelength of that mode, `Lx = 2π/k_max ≈ 14.1`, so the seeded mode fits periodically (replacing the previous broadband white noise on `u`).

## What's unchanged

With `U = h = 1`, `Re = 5×10⁴`, `Pr = 1` the viscosity and diffusivity are still `ν = κ = 2×10⁻⁵`, and the center stratification `N²(0) = 0.1` is unchanged, so the minimum gradient Richardson number is still `0.1`. The grid resolution (128²), 5th-order upwind advection, RK3 time-stepping, diagnostics, and outputs are untouched.

## References

- Kaminski, A. K. & Smyth, W. D. (2019). Stratified shear instability in a field of pre-existing turbulence. *Journal of Fluid Mechanics*, 862, 639–658. https://doi.org/10.1017/jfm.2018.973
- Michalke, A. (1964). On the inviscid instability of the hyperbolic-tangent velocity profile. *Journal of Fluid Mechanics*, 19, 543–556. https://doi.org/10.1017/S0022112064000908


tomchor:
cc @glwagner @navidcy 

I recently did some KHI simulations which led me to this PR, which improves the normalization/dynamics of the KHI example on Oceanostics. Since the Oceanostics example is based on the Oceananigans one, this might be relevant for the Oceananigans docs. 

Long story short is that I wasn't able reproduce the dynamics that appear in previous papers using the normalization and domain used in the Oceananigans docs. I'm only able to reproduce them when applying the changes in this PR. For a visual example, here's what the simulation on Oceanostics `main` looks like (which I think should be the same as in oceananigans): 

https://github.com/user-attachments/assets/c6b69206-cc01-4eab-a75f-51178397c181

and here's what it looks like in this PR:

https://github.com/user-attachments/assets/653a0dd4-190b-4878-9017-f819b4712388

I need to remind myself of the reason the initial Richardson number changed like that (maybe that's the key to it), but based on previous literature we expect secondary KHI to develop along the braids, which does happen in the new version but not on the old one.

Btw, both examples above use the same number of points and nominal Richarson and Reynolds numbers.

It's ofc possible I'm missing something. Curious to hear your thoughts.

glwagner:
seems like the important change is a larger domain in `x`. Makes sense that you get something different if your domain is not a multiple of the most unstable. Honestly, I think you can also miss things by choosing a domain just equal to most unstable (but understand why its necessary for those papers to make their computations feasible).

glwagner:
I think it's welcome to change the Oceananigans example. When you use the power method for real situations you have to choose a very large domain that is many multiples of what you expect the unstable mode to be (because you don't know the answer a priori)

PR #257, opened 2026-06-24
Apply `Filter` boundary behavior via Oceananigans BC types (inherit from field)

### LINKED ISSUE #251: Can we find a better name for `boundary=:edge`?
Currently the options for boundary treatment in filters are:

```
  Stencil offsets that leave the interior 1:Nd_grid of a direction (where Nd_grid is the number of cells along that direction) are handled per-direction. For Periodic directions offsets
  are always wrapped periodically, independent of the boundary keyword. For Bounded directions the boundary keyword picks the policy (default: :shrink):

    •  :shrink — drop out-of-bounds offsets from both the sum and the count, so the filter is an honest local average whose effective stencil shrinks near a wall. This is the default
       for Bounded directions.
    •  :edge — replicate the boundary-cell value (reads ψ[1] or ψ[Nd_grid] for offsets past either end).
    •  (left=a, right=b) — pad with constant a on the low-index side and b on the high-index side (a and b are promoted to a common type).
```

I think `:edge` is the least intuitive option. Maybe we can rename it to something better like `:pad_with_edge` or `:repeat_last_value`?

cc @glwagner

tomchor:
Closes #251

Spatial filters take their boundary treatment from Oceananigans boundary conditions; the symbol API is removed. `boundary_conditions` accepts:

- `nothing` *(default)* — every side inherits the filtered field’s own boundary condition
- a single `BoundaryCondition` — applied to every filtered side
- a `FieldBoundaryConditions` — per side; any unset side inherits from the field

Resolution is per side/direction (explicit overrides, unset inherits), kept separate from the field’s own BCs. Mapping: `ShrinkingBoundaryCondition()` → shrink, zero `Gradient`/`Flux` (incl. `NoFlux`) → edge, `ValueBoundaryCondition(v)` → constant pad. Anything without a discrete analog (non-zero gradient/flux, `Open`, …) falls back to shrinking, so it stays robust across BC kinds. An `AbstractOperation` operand is materialized into a `Field` first (so it too carries conditions to inherit).

```julia
bcs = FieldBoundaryConditions(east=GradientBoundaryCondition(0))  # east set; other sides inherit u’s BCs
ū = GaussianFilter(; dims=(1, 3), σ=0.1, boundary_conditions=bcs)(u)
```

**Breaking:** `:shrink`/`:edge`/`(left=, right=)` and the `boundary` keyword now error with a migration message; the default changes from `:shrink` to inheriting the field’s BCs (a default `CenterField` is `NoFlux` → edge).

### Still open
- Non-zero `Gradient`/`Flux` extrapolation isn’t implemented (falls back to shrinking).
- `show` renders a stored `FieldBoundaryConditions` compactly as `FieldBoundaryConditions(…)`.

### DIFF
--- a/docs/src/filters.md
+++ b/docs/src/filters.md
@@ -69,10 +69,26 @@ julia> size(Field(bf_edge(c))) == size(Field(bf_mixed(c))) == (32, 1, 32)
 true
 ```
 
+#### Oceananigans boundary conditions
+
+The boundary treatment can equivalently be given with Oceananigans' boundary-condition types (passed
+as `boundary` or the alias `boundary_conditions`): [`ShrinkingBoundaryCondition`](@ref) ↔ `:shrink`,
+`GradientBoundaryCondition(0)` (or a zero `FluxBoundaryCondition`) ↔ `:edge`, and
+`ValueBoundaryCondition(v)` ↔ constant padding with `v`. A `FieldBoundaryConditions` sets each side
+of each direction independently — these filter boundary conditions are separate from the field's own.
+
+```jldoctest filters
+julia> bcs = FieldBoundaryConditions(west=ShrinkingBoundaryCondition(), east=GradientBoundaryCondition(0));
+
+julia> size(Field(BoxFilter(; dims=1, N=3, boundary_conditions=bcs)(c)))
+(32, 1, 32)
+```
+
 ### API reference
 
 ```@docs
 BoxFilter
+ShrinkingBoundaryCondition
 ```
 
 ## Gaussian filter
--- a/docs/src/filters.md
+++ b/docs/src/filters.md
@@ -34,7 +34,7 @@ julia> c = CenterField(grid);
 julia> set!(c, (x, z) -> sin(2π * x) * z);
 
 julia> bf = BoxFilter(; dims=(1, 3), N=5)
-BoxFilter(dims=(1, 3), N=5, boundary=:shrink)
+BoxFilter(dims=(1, 3), N=5)
 
 julia> c̄ = Field(bf(c));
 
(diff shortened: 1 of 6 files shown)


tomchor:
@glwagner I put this PR together fully with Claude (so I haven't checked any line of code yet) to have an idea of how it would work if we were to use oceananigans `FieldBoundaryConditions` for filters.

One aspect that makes me reluctant is that `ValueBoundaryCondition` would imply different halo-filling algorithms for a `Field` and for a field-to-be-filtered. For example in Oceananigans a `ValueBC` fills the first halo cell in such a way that an interpolation with the adjacent interior cell implies the current value _exactly_ at the boundary.

For example, for a 1D unity field with a value BC:

```julia
julia> grid = RectilinearGrid(topology=(Bounded, Bounded, Bounded), size=(4, 1, 1), extent=(1, 1, 1));

julia> bcs = FieldBoundaryConditions(grid, (Center(), Center(), Center()); west=GradientBoundaryCondition(0), east=ValueBoundaryCondition(0));

julia> f = CenterField(grid; boundary_conditions=bcs);

julia> set!(f, 1);

julia> Oceananigans.BoundaryConditions.fill_halo_regions!(f);

julia> f.data[:, 1, 1]
10-element OffsetArray(::Vector{Float64}, -2:7) with eltype Float64 with indices -2:7:
  0.0
  0.0
  1.0
  1.0
  1.0
  1.0
  1.0
 -1.0
  0.0
  0.0
```

In the above case the 3 halo cells on the east are `[-1 0 0]`, which is what we'd expect. But same value BC for a filter would imply all halo cells to be `0` (ad infinitum since the filter scale can be made as large as possible).

I think this is at the code of my reluctance to make `ValueBoundaryCondition` be the same as what `boundary = (; left, right)` are doing right now. If you think this is reasonable then I'll accept it, since it does unify user interface.

`GradientBoundaryCondition` seems pretty straightforward to me, as long as we're okay with having the gradient be applied up to infinity instead of only exactly at the boundary.

I don't immediately know what we'd do with `FluxBoundaryCondition`, but maybe let's tackle this after the `ValueBoundaryCondition`...

glwagner:
you may want a different type than `nothing` to indicate "take from `Field`" only because `nothing` is a valid bc option.

glwagner:
> One aspect that makes me reluctant is that ValueBoundaryCondition would imply different halo-filling algorithms for a Field and for a field-to-be-filtered. For example in Oceananigans a ValueBC fills the first halo cell in such a way that an interpolation with the adjacent interior cell implies the current value exactly at the boundary.

I suggest introducing a different name/condition for the different concept / rule that you are trying to express, rather than reusing `ValueBoundaryCondition` for this.

tomchor:
> > One aspect that makes me reluctant is that ValueBoundaryCondition would imply different halo-filling algorithms for a Field and for a field-to-be-filtered. For example in Oceananigans a ValueBC fills the first halo cell in such a way that an interpolation with the adjacent interior cell implies the current value exactly at the boundary.
> 
> I suggest introducing a different name/condition for the different concept / rule that you are trying to express, rather than reusing `ValueBoundaryCondition` for this.

Sorry but I'm kinda lost on what you're proposing. That was the reason why I introduced `boundary=(; left, right)`. But I thought your point was, per https://github.com/tomchor/Oceanostics.jl/issues/251#issuecomment-4700832813, "not to introduce an overlapping concept". But creating something like, say, `ExtendedValueBoundaryConditoin` feels like an overlapping concept. Could you please elaborate, in code, what you are proposing?

Sorry this is moving slow, I'm really having a hard time grasping your point I guess.

tomchor:
Also I guess feel free to just modify this PR to get to where you want if it's easier.

glwagner:
Boundary conditions on `Field` are given by `Field.boundary_conditions`. My point is: use this interface.

This does not mean you cannot introduce new boundary condition classifications. Some time ago we added a new grid, `TripolarGrid`. This grid required a new boundary condition classification, `Zipper()`. Likewise, to express certain concepts with filters could require a new boundary condition type, eg `ExtendedValue()`. Note, `ExtendedValueBoundaryCondition` is a _type alias_ in this philosophy.

The part that is important is that you use `BoundaryConditions` type, `FieldBoundaryConditions` container and also rely on `Field.boundary_conditions`. However, within this philosophy you can still create new boundary condition classifications or change the boundary conditions on a field for filtering purposes.

PR #261, opened 2026-07-07
Add Rayleigh-Taylor instability example (LES with sub-filter-scale KE budget)

tomchor:
## Summary

Adds a documented example, `docs/examples/rayleigh_taylor_instability.jl`, that runs a three-dimensional
Rayleigh-Taylor instability as a large-eddy simulation (LES) and closes the volume-integrated
**sub-filter-scale** kinetic-energy budget with Oceanostics.

Writing it surfaced a performance trap in the spatial filters, so the PR also documents that trap and
adds a live, executed illustration of it to the manual.

## The example

Heavy fluid resting over light fluid (an unstable buoyancy interface at mid-depth) in a `48×24×48`
cuboid, periodic in the two horizontal directions and free-slip in the vertical. There is no surface
forcing: all of the kinetic energy is released from the potential energy stored in the unstable
stratification, with no fluxes through the boundaries.

- LES with a `SmagorinskyLilly(C=0.3)` closure. The large coefficient adds eddy viscosity, which keeps
  this deliberately coarse example well behaved.
- Fourth-order centered advection. A centered scheme is non-dissipative, so essentially all of the
  modeled dissipation comes from the closure and is accounted for explicitly by the budget.
- Coarse-graining uses a horizontal Gaussian filter of width `ℓ = 8Δx`. The vertical, singled out by
  gravity and bounded by the walls, is left unfiltered.

### The budget

Rather than the resolved kinetic energy, the example budgets the energy carried by the scales the filter
removes, `Kˢ = ½τⁱⁱ` with `τⁱʲ = filter(uⁱuʲ) − ūⁱūʲ`:

    d/dt ∫Kˢ dV = ∫Πₖ dV + ∫τ(w,b) dV − ∫εˢ dV

Two sources and one sink:

- `Πₖ` (`KineticEnergyCrossScaleFlux`), the cross-scale flux handed down from the resolved scales. Mind
  the sign: the same term drains the filtered kinetic energy, so what is a sink there is a source here.
- `τ(w,b)` (`subfilter_covariance`), the sub-filter buoyancy flux.
- `εˢ = filter(ε) − ε̄`, assembled from `KineticEnergyDissipationRate` and
  `CoarseGrainedKineticEnergyDissipationRate`.

A hidden `@test` guards closure. It normalizes the residual by the budget's throughput,
`rms(|Πₖ| + |τ(w,b)| + |εˢ|)`, rather than by the smallest term, which is an unstable yardstick because
`Πₖ` passes through zero. The residual measures ~0.8% of throughput against a 2% bound. The docs build
executes the example, so it doubles as a regression test on budget closure.

Physically, the sub-filter scales here are fed mostly by their own buoyancy flux rather than by the
cascade, which is what we should expect: the initial interface is about one grid spacing thick, so most
of the buoyancy variance already sits below the filter scale. The integrated `Πₖ` is even slightly
negative (net backscatter) until roughly four free-fall times, turning into a downscale source only once
the mixing layer is turbulent.

The animation shows buoyancy and `Πₖ`, then the two kinetic energies `Kˡ` and `Kˢ` that the filter
separates, then the volume-integrated budget.

## Filter performance documentation

Oceanostics' separable filters take a fast *staged* path only when the filter operation is the direct
operand of a `Field`. Composing it into another `AbstractOperation` silently falls back to a fused single
kernel, which re-evaluates the filtered operand at every stencil point instead of once per cell. Both
forms return identical values, so the only symptom is speed, and it is severe when the operand is
expensive: `Field(gf(ε)) - ε̄` is roughly five times faster than `gf(ε) - ε̄` when `ε` is
`KineticEnergyDissipationRate`, a nine-term viscous-flux contraction.

- `docs/src/filters.md` gains a `Performance notes` section that demonstrates this with live `@example`
  blocks. It shows the two spellings dispatching to different `compute!` methods, then times both and
  verifies they agree.
- The `BoxFilter` and `GaussianFilter` docstrings point at that section.

Worth knowing for review: those `@example` blocks execute during the docs build and are not cheap, since
they build a small LES and compile the fused kernel (a 15×15 unrolled stencil).

## Changes

- `docs/examples/rayleigh_taylor_instability.jl` (new)
- `docs/make.jl`: register the example
- `docs/src/filters.md`: `Performance notes` section with executed examples, plus an `@id` anchor
- `src/SpatialFilters/box_filter.jl` and `src/SpatialFilters/gaussian_filter.jl`: docstring pointers to it
- `docs/examples/kelvin_helmholtz.jl`: give it a named `@id kelvin_helmholtz_example` anchor so the new
  example's cross-reference resolves, matching the convention the other examples already follow

### DIFF
--- a/docs/examples/kelvin_helmholtz.jl
+++ b/docs/examples/kelvin_helmholtz.jl
@@ -1,4 +1,4 @@
-# # Kelvin-Helmholtz instability
+# # [Kelvin-Helmholtz instability](@id kelvin_helmholtz_example)
 #
 # This example simulates a simple 2D Kelvin-Helmholtz instability and is based on the similar
 # [Oceananigans
--- a/docs/examples/rayleigh_taylor_instability.jl
+++ b/docs/examples/rayleigh_taylor_instability.jl
@@ -247,10 +247,10 @@ wbˢ_pair = @. 0.5 * (∫wbˢ_t[i1] + ∫wbˢ_t[i2])
 
 resid = @. dKˢdt - (Πₖ_pair + wbˢ_pair - εˢ_pair)
 
-using Test                                              #hide
-rms(x) = √(sum(abs2, x) / length(x))                    #hide
-budget_terms = (dKˢdt, Πₖ_pair, wbˢ_pair, εˢ_pair)      #hide
-@test rms(resid) < 0.2 * minimum(rms, budget_terms);    #hide
+using Test                                                          #hide
+rms(x) = √(sum(abs2, x) / length(x))                                #hide
+budget_scale = rms(@. abs(Πₖ_pair) + abs(wbˢ_pair) + abs(εˢ_pair))  #hide
+@test rms(resid) < 0.02 * budget_scale;                             #hide
 
 # ## Plotting
 #
--- a/docs/examples/rayleigh_taylor_instability.jl
+++ b/docs/examples/rayleigh_taylor_instability.jl
@@ -80,7 +80,7 @@ set!(model, b=bᵢ)
 # layer accelerates:
 
 Δx = minimum_xspacing(grid)
-simulation = Simulation(model, Δt = 0.1 * Δx / U, stop_time = 10τ)
+simulation = Simulation(model, Δt = 0.1 * Δx / U, stop_time = 7τ)
 
 conjure_time_step_wizard!(simulation, IterationInterval(5), cfl=0.7, max_change=1.1)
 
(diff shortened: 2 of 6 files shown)


tomchor:
This is what the simulation is looking like so far:

https://github.com/user-attachments/assets/e40c33b3-0bfa-429a-8b00-8015567698ad

tomchor:
https://github.com/user-attachments/assets/3eb4f99b-4934-41c5-879a-bf60bd4ed4a1

Current animation

PR #273, opened 2026-07-19
Lower the Kelvin-Helmholtz example to Re = 2000

tomchor:
Since eaa2147 switched the Kelvin-Helmholtz example from `UpwindBiased(order=5)` to `Centered(order=4)`, the run has been visibly noisier. The scheme change was deliberate and correct for the purpose (implicit upwind dissipation is not a term the filtered KE budget can account for), but it removed the only thing damping the grid scale, and `Re = 5e4` on a 128² grid is far outside what that grid can resolve without it.

## The problem

The sharpest feature in a stratified shear layer is the braid between billows, a sheet that the strain thins until diffusion arrests it at `δ ~ h/√Re`. Resolving that takes a few cells, which ties `Re` to `N`. At `N = 128` over this domain (`Δx ≈ 0.11 h`) the resolvable range is around `Re ~ 10³`, so the example was running roughly an order of magnitude past its grid.

Measured on the existing setup, a sweep over `Re` at fixed `N = 128` gives:

| Re | ν | max\|b\|/B₀ |
|---|---|---|
| 5e4 (before) | 2e-5 | 3.12 |
| 1e4 | 1e-4 | 2.27 |
| 5e3 | 2e-4 | 2.04 |
| 2e3 (this PR) | 5e-4 | 1.16 |
| 1e3 | 1e-3 | 1.000 |
| 5e2 | 2e-3 | 1.000 |

`max|b|/B₀` is an unambiguous error measure here: buoyancy is materially conserved and diffusion only shrinks extrema, so anything above 1 is purely numerical. The example was manufacturing buoyancy extrema more than three times the initial range.

## The change

`Re = 2e3`, which brings the run back into a resolvable range while keeping the billow vigorous (peak available potential energy is about 4.4 against 5.8 at the old value, and `max|w|` about 0.51 against 0.93).

Note that `Re = 2000` still leaves a residual 16% buoyancy overshoot. `Re = 1000` is the first fully clean value, and a survey of published stratified-shear-layer DNS at comparable resolution points lower still, toward `Re ≈ 300`–`600` (Smyth & Moum 2000 ran `Re = 491` at `Δx = 0.109 h`, essentially this grid spacing). `Re = 2000` was chosen deliberately as a compromise that keeps more billow structure; worth knowing it sits at the permissive end.

## Prose fixes

The discussion after the movie still claimed "the upwind scheme here adds some numerical dissipation" and contrasted the example against the centered-advection two-dimensional turbulence example. Both stopped being true in eaa2147. Rewritten to say the budget closes against the explicit `∫εˡ dV` alone.

Also adds a short note at the model setup on why `Re` and `N` are not independent knobs here.

## Testing

Not run on this branch. The embedded budget test (`rms(resid) < 0.06 rms(dKˡdt)`) was verified to pass at `Re = 2000` on a branch with an identical simulation setup, so the docs build should stay green, but CI is the real check.

### DIFF
--- a/docs/examples/kelvin_helmholtz.jl
+++ b/docs/examples/kelvin_helmholtz.jl
@@ -26,7 +26,7 @@ using Oceananigans
 U   = 1     # velocity scale (half the velocity difference across the shear layer)
 h   = 1     # length scale (shear-layer half-width)
 Ri₀ = 0.1   # Richardson number
-Re  = 2e3   # Reynolds number (bounded by the grid; see the resolution note below)
+Re  = 4e3   # Reynolds number (bounded by the grid; see the resolution note below)
 Pr  = 1     # Prandtl number
 
 ν = U * h / Re   # viscosity
@@ -38,7 +38,7 @@ Pr  = 1     # Prandtl number
 # 1964](https://doi.org/10.1017/S0022112064000908)), so that the perturbation we seed below fits
 # periodically:
 
-N = 128
+N = 256
 k_max = 0.4446 / h   # most unstable KH wavenumber (Michalke, 1964)
 Lx = 2π / k_max      # one most-unstable wavelength
 Lz = 10
@@ -84,7 +84,7 @@ set!(model, u=shear_flow, b=stratification, w=perturbation)
 # adapts it as the flow evolves:
 
 Δx = minimum_xspacing(grid)
-simulation = Simulation(model, Δt = 0.1 * Δx / U, stop_time=120)
+simulation = Simulation(model, Δt = 0.2 * Δx / U, stop_time=120)
 conjure_time_step_wizard!(simulation, IterationInterval(2), cfl=0.8, max_Δt=1)
 
 # ## Model diagnostics
--- a/docs/examples/kelvin_helmholtz.jl
+++ b/docs/examples/kelvin_helmholtz.jl
@@ -49,15 +49,6 @@ model = NonhydrostaticModel(grid; timestepper = :RungeKutta3,
                             closure = ScalarDiffusivity(; ν, κ),
                             buoyancy = BuoyancyTracer(), tracers = :b)
 
-# The centered scheme is what makes the budget below meaningful: it carries no numerical dissipation, so
-# every sink in the kinetic-energy budget is one we compute explicitly. The price is that nothing damps
-# the grid scale for us, which is what caps `Re` at the value set above. In a stratified shear layer the
-# sharpest feature is the braid between billows, a sheet the strain thins until diffusion arrests it at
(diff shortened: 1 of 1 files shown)


docs/examples/kelvin_helmholtz.jl
```diff
@@ -49,6 +49,15 @@ model = NonhydrostaticModel(grid; timestepper = :RungeKutta3,
...
 
+# The centered scheme is what makes the budget below meaningful: it carries no numerical dissipation, so
+# every sink in the kinetic-energy budget is one we compute explicitly. The price is that nothing damps
+# the grid scale for us, which is what caps `Re` at the value set above. In a stratified shear layer the
+# sharpest feature is the braid between billows, a sheet the strain thins until diffusion arrests it at
+# `δ ∼ h / √Re`; resolving it takes a few cells, so `Re` and `N` are not independent knobs. At `N = 128`
+# over this domain (`Δx ≈ 0.11 h`), that lands near `Re ∼ 10³`. Raising `Re` without also raising `N`
+# leaves the braid unresolved, and it then shows up in the budget as grid-scale noise: at `Re = 5×10⁴`
+# on this grid the billow manufactures buoyancy extrema roughly three times the initial range `±B₀`.
+#
```
tomchor:
```suggestion
```

docs/examples/kelvin_helmholtz.jl
```diff
@@ -309,7 +314,7 @@ end
...
 # feeds the subfilter scales through the cross-scale flux (`−∫Πₖ dV`), while the coarse-grained viscous
 # dissipation `∫εˡ dV` stays comparatively small at this Reynolds number. The residual (dashed), the
-# sum of the negative tendency `−d(∫Kˡ)/dt` and the three source terms, stays small. Unlike the centered-advection
-# [Two-dimensional turbulence example](@ref two_d_turbulence_example), the upwind scheme here adds some
-# numerical dissipation, but because it acts mostly at the grid scale it barely projects onto the
-# smooth filtered budget, which still closes to within a few percent.
+# sum of the negative tendency `−d(∫Kˡ)/dt` and the three source terms, stays small. As in the
+# [Two-dimensional turbulence example](@ref two_d_turbulence_example), the centered scheme contributes no
+# numerical dissipation of its own, so the budget closes against the explicit `∫εˡ dV` alone rather than
+# against a scheme-dependent sink we would have no way to measure.
```
tomchor:
```suggestion
# [Two-dimensional turbulence example](@ref two_d_turbulence_example), the centered scheme contributes no
# numerical dissipation of its own, so the budget closes against the explicit `∫εˡ dV` alone with a negligible residual.
```

PR #276, opened 2026-07-26
Local APE budget: Υ, ε_A and the diffusive buoyancy flux

tomchor:
Three new diagnostics of the local available potential energy budget, plus the naming and documentation work that fell out of placing them.

## New diagnostics

```
Υ   = z✶ - z          [m]        BuoyancyDisplacementPotential            AvailablePotentialEnergyEquation
ε_A = κ ∂ᵢb ∂ᵢΥ       [m² s⁻³]   AvailablePotentialEnergyDissipationRate  AvailablePotentialEnergyEquation
Φ   = κ ∂b/∂z = -q₃   [m² s⁻³]   PotentialEnergyDiffusiveBuoyancyFlux     PotentialEnergyEquation
```

Υ is `∂e_a/∂b`, so contracting it with `∇b` gives ε_A and contracting it with a sub-filter buoyancy flux gives a cross-scale APE flux. ε_A is the sink of the local APE equation of [Wenegrat, Chor & Barkan (2026)](https://arxiv.org/abs/2605.15879): the Winters diapycnal mixing rate less Φ, the diffusion the reference state undergoes on its own. Splitting Φ out makes that decomposition checkable rather than asserted, `d/dt ∫e_b dV = ∫(ε_A + Φ) dV ≥ 0`, with neither term sign-definite alone.

Υ and ε_A take `AvailablePotentialEnergy`'s `(model; method)` / `(model, z✶)` interface so a set of diagnostics can share one sort, and ε_A also takes `upsilon`. `κ∇b` comes from the closure's own `diffusive_flux_*`, which restricts ε_A and Φ to `BuoyancyTracer` (enforced).

Φ lives in `PotentialEnergyEquation` rather than with ε_A: it needs no reference state, and it is the diffusive conversion term of the `e_p` equation before it is anything to do with `e_a`. The APE module re-exports it. Its short name `DiffusiveBuoyancyFlux` is module-scoped; the prefixed alias is what `using Oceanostics` gets, which keeps it clear of `TracerZDiffusiveFlux`. Those two are genuinely different objects from the same closure call: the tracer diagnostic is `q₃`, down-gradient and on the `z` face, while this is `-q₃` at cell center, the sign and location the energy budgets want.

## Breaking: `wb` renamed

`uᵢbᵢ` is the one term the kinetic and potential energy budgets share, and each side reads it differently, so it now carries a name per point of view. Each module names the *other* reservoir:

| Name | Exported from |
|---|---|
| `PotentialToKineticEnergyConversion` | `Oceanostics` and all three modules |
| `KineticEnergyEquation.PotentialEnergyConversion` | that module only |
| `PotentialEnergyEquation.KineticEnergyConversion`, likewise APE | those two modules only |

This replaces **`KineticEnergyBuoyancyProduction`** (the old top-level export) and **`BuoyancyProduction`** (the old local name). Both are gone, so this is a breaking change for anyone using either. The `show` description changed with it.

## Documentation

The potential energy page now derives its equation from the buoyancy tracer equation and tabulates every term against whether a diagnostic exists for it. Only `e_p` and Φ are implemented there; the buoyancy conversion is borrowed from the KE module, and advection, transport, tendency and forcing have no diagnostic yet.

Notation is now uniform: lower-case `e` for an energy density, the pointwise quantity a diagnostic returns, and upper-case `E` for its volume integral, so `E_p = ∫e_p dV = Integral(PotentialEnergy(model))`.

The lock release example closes both halves of the exchange,

```
d/dt ∫KE dV =  ∫wb dV − ∫ε   dV
d/dt ∫e_a dV = −∫wb dV − ∫ε_A dV
```

to 0.07% of the KE tendency and 0.6% of ∫ε_A, the smallest term in the APE budget. That needed the example off `WENO` and onto `Centered(order=4)` with an explicit closure, since a dissipative advection scheme puts the sinks where no diagnostic can see them.

Also fixes a cross-reference that was breaking the docs assemble: the ε_A docstring pointed at `TracerVarianceEquation.DissipationRate`, but that module exports both the short and prefixed names and Documenter keys `@ref` on the binding, so the short-name target had no docstring to resolve against.

## Testing

Handed `b` in place of Υ, the ε_A kernel is exactly `χ/2`, so it is checked against `TracerVarianceDissipationRate` to roundoff. Φ telescopes exactly to `κA[b(z_top) − b(z_bottom)]`, which pins its boundary treatment, and `ε_A + Φ` is checked non-negative cell by cell and in the integral on both regular and stretched grids. `test_pe_diagnostics.jl` also asserts the export surface above, including the negatives, since nothing in the code enforces it.

`ape_diagnostics` 388/388, `pe_diagnostics` 114/114, `perf_invariants` 80/80 as of the last full run; the final `BuoyancyProduction` rename was pushed without rerunning them, so `ke_diagnostics` is worth a look. Doctests were replayed against real `show` output at each rename, and every qualified `@ref` in the package was swept against every `@docs` entry.

ε_A and Υ are already in use in a Kelvin-Helmholtz APE study, where ε_A was validated against an independent offline (Python) implementation.


src/PotentialEnergyEquation.jl
```diff
@@ -187,4 +204,121 @@ end
...
+#---
+
+#+++ Diffusive buoyancy flux
+# `Φ = κ ∂b/∂z = -q₃`, the vertical diffusive buoyancy flux taken upward. It is the diffusive conversion
+# term of the `e_p` equation, the only way diffusion changes the potential energy of a closed domain,
+# and it is also the second of the two parts `ε_A` is written out of. It comes off the closure's own
+# `diffusive_flux_z`, exactly as `ε_A` does, so the two always carry the same `κ`. The flux lives on the
+# `z` face, so it is interpolated to the cell center.
+@inline diffusive_buoyancy_flux_ccc(i, j, k, grid, args...) =
+    -ℑzᵃᵃᶜ(i, j, k, grid, diffusive_flux_z, args...)
```
tomchor:
```suggestion
@inline diffusive_buoyancy_flux_ccc(i, j, k, grid, args...) = -ℑzᵃᵃᶜ(i, j, k, grid, diffusive_flux_z, args...)
```

PR #281, opened 2026-08-03
More unicode integrals

tomchor:


tomchor:
Setting up `NetCDFWriter`s with the unicode version of the integrals produces a `Field`. This can be useful for usage in operations, but apparently makes the writing a bit slower. So I need to think about whether the added user-friendliness is worth it

PR #287, opened 2026-08-09
Use DocumenterCodeBlocks for nicer code blocks in the docs

tomchor:
Adds [DocumenterCodeBlocks.jl](https://github.com/fredrikekre/DocumenterCodeBlocks.jl).

The `plugins = [CodeBlocks()]` to `makedocs` tokenizes Julia blocks at build time with JuliaSyntax instead of leaving them to highlight.js in the browser, which at its defaults gives:

- line numbers, with clickable gutters producing stable URL fragments like `#c-e61e7f97-L3-L7`,
- better colors, since a real parse can tell function calls, types and macros apart,
- identifiers linked to their docstrings, with hover popups showing the signature.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -4,6 +4,7 @@ using Pkg; Pkg.instantiate()
 using Base64
 
 using Documenter
+using DocumenterCodeBlocks: CodeBlocks
 using Literate
 
 using Oceananigans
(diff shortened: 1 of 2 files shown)


docs/make.jl
```diff
@@ -165,13 +166,19 @@ format = Documenter.HTML(collapselevel = 1,
 #---
 
 #+++ Make the docs
+# `CodeBlocks()` tokenizes Julia code blocks at build time with JuliaSyntax instead of leaving
+# them to highlight.js in the browser. That buys a real parse (so function calls, types and
+# macros get their own colors), line numbers with linkable gutters, and hover popups linking
+# identifiers to their docstrings. It only touches `julia`/`julia-repl`/`jldoctest` blocks, so
+# the executed-output blocks retagged by `colorize_ansi_output` above are left alone.
```
tomchor:
```suggestion
```

Issue #206, opened 2025-06-08
Package notation differs from Oceananigans


glwagner:
When reading a few docstrings, I noticed that a different notation is used from the notation that is used in the Oceananigans source code in some places. For example, the "nonhydrostatic pressure anomaly" is called `pₙₕₛ` here:

https://github.com/tomchor/Oceanostics.jl/blob/5512355a520fa979426180a786e656c2f8cdce87/src/TKEBudgetTerms.jl#L114

but in Oceananigans it is called `pNHS`:

https://github.com/CliMA/Oceananigans.jl/blob/51a77cc6ab9d7f66eb84ed574709adac3c9210c9/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl#L213

Notation in Oceananigans is strict/meaningful --- for example, `set!(model, c=...)` requires the kwarg and tracer name to match. There should be no exceptions to this (or if there are, its a bug).

I do like `pₙₕₛ` (I think `pₙ` would be even better) , and support changing Oceananigans to match the notation here. Either way one has to change.

glwagner:
That's a good point that `pₙ` is easily confused with "pressure at time-step `n`". It's still better the pNHS though. We could also have `nh` and `h`

glwagner:
Single letter subscripts are and will forever be better than multi-letter subscripts, so nothing has changed...

tomchor:
> Single letter subscripts are and will forever be better than multi-letter subscripts, so nothing has changed...

...okay. But what does this mean for the issue at hand? Are you okay now changing the notation everywhere to `pₙ` and `pₙₕ`?

tomchor:
> I do like `pₙₕₛ` (I think `pₙ` would be even better) , and support changing Oceananigans to match the notation here. Either way one has to change.

This part of the top post led me to believe you wanted to solve this here. But okay, I'll wait for some action on the Oceananigans side

PR #308, opened 2026-09-02
RichadsonNumber -> GradientRichardsonNumber

### LINKED ISSUE #304: Rename `RichardsonNumber` to `GradientRichardsonNumber`
`RichardsonNumber` computes the *gradient* Richardson number, `(∂b/∂z) / |∂u⃗ₕ/∂z|²`. Since bulk and flux Richardson numbers also exist, the name should say which one it is. Suggest renaming to `GradientRichardsonNumber`, keeping `RichardsonNumber` as an alias.

tomchor:
Closes https://github.com/tomchor/Oceanostics.jl/issues/304

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oceanostics"
 uuid = "d0ccf422-c8fb-49b5-a76d-74acdde946ac"
-version = "0.20.2"
+version = "0.21.0"
 authors = ["tomchor <tomaschor@gmail.com>"]
 
 [deps]
--- a/docs/examples/kelvin_helmholtz.jl
+++ b/docs/examples/kelvin_helmholtz.jl
@@ -89,9 +89,9 @@ progress = ProgressMessengers.TimedMessenger()
 simulation.callbacks[:progress] = Callback(progress, IterationInterval(200))
 
 
-# We can also define some useful diagnostics of the flow, starting with the `RichardsonNumber`
+# We can also define some useful diagnostics of the flow, starting with the `GradientRichardsonNumber`
 
-Ri = RichardsonNumber(model)
+Ri = GradientRichardsonNumber(model)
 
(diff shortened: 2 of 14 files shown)


tomchor:
Pushed b9448d2, which fixes the shear term so the new name is backed by the implementation.

## What was wrong

`richardson_number_ccf` built its denominator by differentiating the speed and then squaring, `(∂|u⃗ₕ|/∂z)²`. The gradient Richardson number needs `|∂u⃗ₕ/∂z|²`, the squared norm of the shear vector. Differentiation and the norm do not commute, and by Cauchy-Schwarz the computed denominator is the smaller of the two, so the diagnostic overestimated Ri whenever the shear vector turned with height. The docstring already stated the correct formula, so the code and the documentation disagreed.

The existing test used `u = S*z + S*y` with `v = 0`, which is unidirectional, which is why this survived. The Kelvin-Helmholtz example is unidirectional too.

## What changed

The shear vector is now built first: each velocity component is differentiated along the true vertical, and the part of the resulting vector lying along that vertical is removed, leaving the shear of the true-horizontal velocity. For a `NegativeZDirection` gravity this reduces to `(∂u/∂z)² + (∂v/∂z)²`, which matches Oceananigans' own `shear_squaredᶜᶜᶠ` in `ri_based_vertical_diffusivity.jl`, so this also closes a small piece of #206.

Following a tilted gravity needs all nine velocity-gradient components at `(Center, Center, Face)`, so there are three new directional-derivative helpers. They replace `uₕ_norm_ccc` and `w²_from_u⃗_tilted_ccc`, which no longer have callers, along with the module-local `ψ²` they used.

One discretization note: the components are interpolated and then squared, rather than squared and then interpolated the way Oceananigans does it. The tilted path forces this, since projecting out the vertical part requires the components at a common location, and it is what makes the cancellation exact. The two differ only at truncation error.

tomchor:
@copilot resolve the merge conflicts in this pull request

